Issue #82, opened 2017-01-06
Detect when disconnect from Master


PerrineAguiar:
The app has no way to know if the Master is disconnected while running right now, so if the master is shut down during the operation, the light won't change it's color to red.
Also, there is no way of reconnecting to the master after a disconnection but restarting the app.

PR #104, opened 2017-01-12
Feature/enable restart with snackbar

### LINKED ISSUE #67: Avoid closing app to restart node and apply new settings
One possibility is to detect that settings changed and display a dialog which asks the user if he wants to restart the node/app.

PerrineAguiar:
This PR adds a snackbar to allow the user to restart the running activity if the ros master settings changed.

screenshot_2017-01-12-17-29-26

Fixes #67.


smits:
Can we move forward on this one?  Would be nice to have it in.

PerrineAguiar:
@smits Can we close this?
With #275 we don't need to restart the app if ros connection failed. A snackbar appears with the reconnect button in the RunningActivity instead.

Note: You still need to restart the app if you change the ros master settings while the nodes are running, but that seems ok, it is the same behavior when running a node on linux.

smits:
I would prefer a possibility to reconnect to a master without having to restart the app, but I understand that it might be hard to accomplish.

PerrineAguiar:
Ok, I will take a closer look at it when tackling #82.

PR #130, opened 2017-01-25
Feature/get pose in area description

### LINKED ISSUE #11: Allow user to get device pose in area description frame
e.g. TANGO_COORDINATE_FRAME_AREA_LEARNING

Currently the device pose is available only in TANGO_COORDINATE_FRAME_START_OF_SERVICE.

PerrineAguiar:
This PR adds a transform to /tf: area_description_T_start_of_service

Fixes #11.

### DIFF
--- a/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h
+++ b/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h
@@ -137,6 +137,7 @@ class TangoRosNode {
 
   tf::TransformBroadcaster tf_broadcaster_;
   geometry_msgs::TransformStamped start_of_service_T_device_;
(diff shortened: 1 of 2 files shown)


PerrineAguiar:
@jubeira @mcopejans ptal

jubeira:
The code looks good @PerrineAguiar .
One question: wouldn't it be nice to be able to configure if you want both poses or not? As I understand, this new frame is used for corrections in some cases, so from a client point of view, I don't know if both of them will be required in all cases.

PerrineAguiar:
@jubeira Thanks for your review.
The area_description frame is available only if Tango drift correction (or area learning) is enable. So the user can already configure this. However I agree that this is not clear, especially for users not knowing Tango. I made an issue #142 to update the documentation.

PR #132, opened 2017-01-26
Feature/run roscore in app

### LINKED ISSUE #127: Allow to run roscore inside the app

PerrineAguiar:
This PR allows to run the ROS master on the device.

Fixes #127.


PerrineAguiar:
@jubeira Thanks for your review.
It was also showing this kind of weird URI, this is the URI returned by getMasterURI. I tried to connect using this URI and it was working perfectly.
However it can be confusing, so I changed the code to show the IP address of the device.

PR #133, opened 2017-01-26
Fix/app crashes when no network

### LINKED ISSUE #123: App crashes if there is not network connection.
**Observed behavior**
When starting the app if there is not wifi connection the app crashes

**Desired behavior**
The app should not crash and just mention that the ROS master is not reachable or something similar.

PerrineAguiar:
This PR  fixes #123.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
@@ -300,49 +300,49 @@ public class RunningActivity extends AppCompatRosActivity implements TangoRosNod
 
     @Override
     protected void init(NodeMainExecutor nodeMainExecutor) {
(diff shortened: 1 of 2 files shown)


jubeira:
LGTM @PerrineAguiar.
I think that this could be solved using a smaller try-catch block. I just tested it here, and I believe the network error arises when getting the host address, so something like 

`        NodeConfiguration nodeConfiguration;

        try {

            nodeConfiguration = NodeConfiguration.newPublic(InetAddressFactory.newNonLoopback().getHostAddress());
            nodeConfiguration.setMasterUri(this.nodeMainExecutorService.getMasterUri());

        } catch (RosRuntimeException e) {
            Log.e(TAG, getString(R.string.network_error));
            displayToastMessage(R.string.network_error);
            return;
        }`

at the beginning of the `init` method also should work. As in this case you return when you get the error, I think it's a bit more readable with the smaller try catch block.

PerrineAguiar:
@jubeira Thanks for your review. I addressed your comment, you are right it is more readable.

PR #136, opened 2017-01-26
Catch connection errors happening when using usb tethering

### LINKED ISSUE #134: Connection failure not caught with usb tethering

PerrineAguiar:
This PR fixes #134.

### DIFF
--- a/TangoRosStreamer/tango_ros_node/src/main/java/eu/intermodalics/tango_ros_node/TangoRosNode.java
+++ b/TangoRosStreamer/tango_ros_node/src/main/java/eu/intermodalics/tango_ros_node/TangoRosNode.java
@@ -33,6 +33,8 @@ import org.ros.node.Node;
 public class TangoRosNode extends NativeNodeMainBeta {
     public static final int ROS_CONNECTION_ERROR = 1;
     public static final String ROS_CONNECTION_FAILURE_ERROR_MSG = "ECONNREFUSED";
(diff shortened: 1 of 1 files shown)


TangoRosStreamer/tango_ros_node/src/main/java/eu/intermodalics/tango_ros_node/TangoRosNode.java
```diff
@@ -70,12 +80,12 @@ public void attachCallbackListener(CallbackListener callbackListener) {
...
-        if (throwable.getMessage().contains(ROS_CONNECTION_FAILURE_ERROR_MSG) ||
-                throwable.getMessage().contains(ROS_CONNECTION_UNREACHABLE_ERROR_MSG) ||
-                throwable.getMessage().contains(ROS_CONNECTION_TIMEOUT_ERROR_MSG) ||
-                throwable.getMessage().contains(ROS_WRONG_HOST_NAME_ERROR_MSG)) {
-            executeOnErrorHook(ROS_CONNECTION_ERROR);
-        } else if (this.executeReturnCode != 0) {
+        for (String errorMessage : errorMessages) {
+            if (throwable.getMessage().contains(errorMessage)) {
+                executeOnErrorHook(ROS_CONNECTION_ERROR);
+            }
```
jubeira: nit: perhaps I would add a a `return` statement as the error hook should be executed only once, so there's no need to keep iterating.

PR #137, opened 2017-01-30
Feature/log to rosout

### LINKED ISSUE #59: Forward Tango Error/Warning/Info logcat messages to /rosout

PerrineAguiar:
This PR add logs to rosout.

* On the Java side, RosoutLogger is used. It logs to logcat and publishes the message to rosout with the correct log level.
* On the C++ side, a wrapper around ros log and glog is used.

Fixes #59.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
@@ -47,8 +47,6 @@ import eu.intermodalics.tango_ros_node.TangoRosNode;
  * and to update Dynamic Reconfigure with changes applied on the app's preferences by the user.
  */
 public class ParameterNode extends AbstractNodeMain implements NodeMain, SharedPreferences.OnSharedPreferenceChangeListener {
-    private static final String TAG = ParameterNode.class.getSimpleName();
-
     private static final String NODE_NAME = "parameter_node";
     private static final String RECONFIGURE_TOPIC_NAME = "parameter_updates";
     private static final String RECONFIGURE_SRV_NAME = "set_parameters";
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
@@ -20,6 +20,7 @@ import android.app.Activity;
 import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
 
+import org.apache.commons.logging.Log;
 import org.ros.exception.RemoteException;
 import org.ros.exception.ServiceNotFoundException;
 import org.ros.message.MessageListener;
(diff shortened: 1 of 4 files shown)


tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -244,16 +279,18 @@ TangoErrorType TangoRosNode::TangoSetupConfig() {
 
   tango_config_ = TangoService_getConfig(TANGO_CONFIG_DEFAULT);
   if (tango_config_ == nullptr) {
-    LOG(ERROR) << function_name << ", TangoService_getConfig error.";
+    Log(ERROR, std::ostringstream().flush() <<
```
adolfo-rt: Why not flush in the wrapper to avoid repetition?

PerrineAguiar:
@jubeira We log to rosout only the messages from the Parameter node and the native TangoRosNode, so it is normal that there is less messages in rosout than in logcat. Moreover, it is mostly error messages, so if there is no errors, there will be no messages log to rosout.
Concerning the "ROS MASTER IS UP" message, I think the reason why it is not logged is that the node is not fully started at that time; according to the [ros wiki](http://wiki.ros.org/roscpp/Overview/Logging#Output): _Note that until the node is fully started, messages will not be sent, so you may not see initial messages_.

Concerning the RunningActivity messages, the issue is that we need a node connected to ROS to be able to publish to rosout. We could use the Parameter node or have a third node publishing RunningActivity messages to rosout. However, since the RunningActivity messages are mostly errors about not being able to connect to ROS, these messages won't even be published to rosout. The only messages we might miss are some tango errors.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: That would be best indeed, maybe I am missing something but I don't see how.
I am using flush to create a temporary `std::ostream&` that I need to make messages using the `<<` operator. 
More info here: http://stackoverflow.com/questions/8287188/stdostringstream-printing-the-address-of-the-c-string-instead-of-its-content

Another possibility would be to use `std::string` instead of `std::ostream`, using strings concatenation with the `+` operator. But that would require to convert any non-string value to string.

jubeira:
I see, thanks for the explanation @PerrineAguiar !

PerrineAguiar:
@smits Should I merge this PR or do you want to take a look and make the native wrapper more proper (so it can be a contribution for ros_android_native)?

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
```diff
@@ -172,20 +169,20 @@ private void callDynamicReconfigure(String paramName, boolean paramValue) {
             serviceClient.call(srv_req, new ServiceResponseListener<ReconfigureResponse>() {
                 @Override
                 public void onSuccess(ReconfigureResponse reconfigureResponse) {
-                    Log.i(TAG, "Dynamic Reconfigure success");
+                    mConnectedNode.getLog().info("Dynamic Reconfigure success");
```
smits: Could we store the log at the beginning of the function and then reuse this, e.g. add  `Log log = mConnectedNode.getLog();` somewhere at L159

PerrineAguiar:
@smits Yes, but the idea was to log both on logcat and on rosout.

PerrineAguiar:
It was decided not to use the wrapper on the native side. The reason is that the message logged on the native side are mostly tango related and not ros related.
The PR #141 is still there if we want to tackle this later by making a proper wrapper.

PR #138, opened 2017-02-01
Feature/point cloud as laser scan

### LINKED ISSUE #126: Publish flattened 3d point cloud as laser scan

PerrineAguiar:
This PR allows publishing a laser scan message computed from the tango point cloud.
The method to convert the point cloud to a laser scan is taken from the ros package [point_cloud_to_laser_scan](http://wiki.ros.org/pointcloud_to_laserscan). 

Fixes #126.

### DIFF
--- a/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
+++ b/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
@@ -21,10 +21,8 @@
 
 #include <dynamic_reconfigure/config_tools.h>
 #include <dynamic_reconfigure/server.h>
-#include <sensor_msgs/LaserScan.h>
 #include <sensor_msgs/PointField.h>
 #include <sensor_msgs/point_cloud2_iterator.h>
-//#include <tf2_ros/buffer.h>
 
 namespace {
 // This function routes onPoseAvailable callback to the application object for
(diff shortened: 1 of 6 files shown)


tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -105,13 +105,130 @@ void toPointCloud2(const TangoPointCloud& tango_point_cloud,
...
+
+    // Laser scan frame is rotated of 90 degrees around x axis with respect to
+    // point cloud frame.
+    double x = tango_point_cloud.points[i][0];
+    double y = tango_point_cloud.points[i][2];
+    double z = -tango_point_cloud.points[i][1];
+    toLaserScanRange(x, y, z, min_height, max_height, laser_scan);
+  }
+  point_cloud->header.stamp.fromSec((tango_point_cloud.timestamp + time_offset) / 1e3);
+  laser_scan->header = point_cloud->header;
```
mcopejans: The code above seems dupliate to toLaserScan(point_cloud). Why not use this function?

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -404,7 +542,33 @@ void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
...
+                                     / laser_scan_.angle_increment);
+    // Laser scan rays with no obstacle data will evaluate to infinity.
+    laser_scan_.ranges.assign(ranges_size, std::numeric_limits<double>::infinity());
+    if (publisher_config_.publish_point_cloud && point_cloud_available_mutex_.try_lock()) {
+      toPointCloud2AndLaserScan(*point_cloud, time_offset_, laser_scan_min_height_,
+                               laser_scan_max_height_, &point_cloud_, &laser_scan_);
+      point_cloud_.header.frame_id = toFrameId(TANGO_COORDINATE_FRAME_CAMERA_DEPTH);
+      point_cloud_available_.notify_all();
+      point_cloud_available_mutex_.unlock();
+    } else {
```
mcopejans: It looks like the laser scan is always published when the point cloud is published.

Why not simply do:
```
if (publish_point_cloud) toPointCloud2 ...
if (publish_laser_scan) toLaserScan() ...
```

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
jubeira: Hmm perhaps it is because iterating over the same point cloud twice is expensive, so in this way both tasks are accomplished in the same iteration. Am I right?
If that is the case, maybe I would write the initialization part of `toPointCloud2` in a smalll separate function to at least avoid repeating that code in `toPointCloud2` and `toPointCloud2AndLaserScan`

mcopejans: It is still the same order of magnitude and would make the code more readable.

jubeira: Yes, I certainly agree on that one (and it would be more maintainable too).
Even being the same order of magnitude, I don't know for sure if in this case it would affect performance to iterate twice. Pointclouds are pretty heavy as I understand. 

If iterating twice affected performance in a non-negligible way, it wouldn't be a bad idea to me to have two separate functions with their respective `for` loops as it is now. If the performance cost is negligible, I totally agree with you @mcopejans .

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
jubeira: I believe this is related to having a function that publishes both at the same time. 
The else clause below publishes the laser scan alone. This part will be resolved once a decision is taken about the function `toPointCloud2AndLaserScan`

PerrineAguiar:
@mcopejans @jubeira Thanks for your review.
After checking that it does not affect performance, I removed the toPointCloud2AndLaserScan function. As you pointed out it makes the code more readable and avoid code duplication.

PR #152, opened 2017-02-13
Feature: wrap application as a Catkin package

jubeira:
This PR wraps the application inside a Catkin package.
Why is this useful/ important?
-The app can now be built running `catkin_make` from the root workspace folder after sourcing ROS kinetic environment. This command will create the apk file too.
-Using catkin_make, the app and the node's artifacts are uploaded to the local workspace Maven repository (tango_ros_ws/devel/share/maven). This allows using the code for the `TangoRosNode` by sourcing the app's workspace after running catkin_make (`source tango_ros_ws/devel/setup.bash`).

Note 1: OpenCV version was changed to 3 in `tango_ros_native` CMakelists'. OpenCV 3 is the [default for ROS kinetic](http://wiki.ros.org/opencv3). This doesn't seem to cause problems. See #102 .
Note 2: When running catkin_make the first time, glog (third party dependency) downloads some code required to build. Apparently there is a race condition and catkin keeps on building the application before the download is finished, so the first time it fails. It doesn't happen running `catkin_make -j1`.
Note 3: even though `catkin_make` and `catkin_make_isolated` work, I get an error with `catkin build` when the android NDK tries to build the miniglog. I couldn't figure out what is happening here; @PerrineAguiar would you please take a look?

### DIFF
--- a/TangoRosStreamer/CMakeLists.txt
+++ b/TangoRosStreamer/CMakeLists.txt
@@ -1,0 +1,25 @@
+##############################################################################
+# CMake
+##############################################################################
+
+cmake_minimum_required(VERSION 2.8.3)
+project(tango_ros_streamer)
+
+##############################################################################
+# Catkin
+##############################################################################
+
+find_package(catkin REQUIRED rosjava_build_tools)
+# Set the gradle targets you want catkin's make to run by default
+# e.g. usually catkin_android_setup(assembleRelease uploadArchives)
+catkin_android_setup(assembleRelease assembleDebug uploadArchives)
+catkin_package()
+
+
+##############################################################################
+# Installation
+##############################################################################
+
+# Deploy android libraries (.aar's) and applications (.apk's)
+install(DIRECTORY ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_MAVEN_DESTINATION}/com/github/rosjava/${PROJECT_NAME}/ 
+       DESTINATION ${CATKIN_GLOBAL_MAVEN_DESTINATION}/com/github/rosjava/${PROJECT_NAME}/)
--- a/TangoRosStreamer/build.gradle
+++ b/TangoRosStreamer/build.gradle
@@ -22,6 +22,14 @@ buildscript {
     apply from: "https://github.com/rosjava/android_core/raw/kinetic/buildscript.gradle"
 }
 
+apply plugin: 'catkin'
+
+allprojects {
+    /* A github url provides a good standard unique name for your project */
+    group 'eu.intermodalics'
+    version = project.catkin.pkg.version
+}
+
 subprojects {
     apply plugin: 'ros-android'
 
--- a/TangoRosStreamer/package.xml
+++ b/TangoRosStreamer/package.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package>
   <name>tango_ros_streamer</name>
-  <version>0.1.0</version>
+  <version>1.0.13</version>
   <description>This package wraps Tango Ros Streamer application</description>
 
   <maintainer email="ruben@intermodalics.eu">Ruben Smits</maintainer>
--- a/TangoRosStreamer/package.xml
+++ b/TangoRosStreamer/package.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0"?>
+<package>
+  <name>tango_ros_streamer</name>
+  <version>0.1.0</version>
+  <description>This package wraps Tango Ros Streamer application</description>
+
+  <maintainer email="ruben@intermodalics.eu">Ruben Smits</maintainer>
+  <maintainer email="perrine@intermodalics.eu">Perrine Aguiar</maintainer>
+  <license>Apache</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>tango_ros_native</build_depend>
+  <build_depend>rosjava_build_tools</build_depend>
+
+</package>
--- a/TangoRosStreamer/tango_ros_node/build.gradle
+++ b/TangoRosStreamer/tango_ros_node/build.gradle
@@ -59,7 +59,7 @@ task ndkBuild(type: Exec) {
 
 dependencies {
     // e.g. official msgs
-    compile('org.ros.rosjava_core:rosjava:[0.3.0]') {
+    compile('org.ros.rosjava_core:rosjava:[0.3, 0.4)') {
         exclude group: 'junit'
         exclude group: 'xml-apis'
     }
--- a/tango_ros_common/tango_ros_native/CMakeLists.txt
+++ b/tango_ros_common/tango_ros_native/CMakeLists.txt
@@ -14,7 +14,7 @@ find_package(catkin REQUIRED COMPONENTS
    )
 
 find_package(cmake_modules REQUIRED)
-find_package(OpenCV 2 REQUIRED)
+find_package(OpenCV 3 REQUIRED)
 
 generate_dynamic_reconfigure_options(
   cfg/Publisher.cfg


PerrineAguiar:
@jubeira Thanks for working on this!

I noticed two issues:
* My system has ros indigo installed, so OpenCV 3 is not installed. I am not sure what the best solution is: force people to install OpenCV 3 or support both cases?
* Java8 needs to be installed and JAVA_HOME needs to be set to "/usr/lib/jvm/java-8-oracle" to be able to build the .apk

Finally, the installation instructions will have to be updated. I can take care of it once this PR is merged.

adolfo-rt:
> Note 2: When running catkin_make the first time, glog (third party dependency) downloads some code required to build. Apparently there is a race condition and catkin keeps on building the application before the download is finished, so the first time it fails. 

Sounds like a missing package/target dependency. Where/how is the glog dependency specified?

jubeira:
@PerrineAguiar thanks for your quick reply. I'm not sure what's the best solution for this either. Perhaps @smits or @adamantivm have some opinion about it. And yes, the install instructions should be modified, but perhaps it would be nice to make this work with `catkin build` too.

@adolfo-rt this is what I know about it: `tango_ros_native` is a catkin package, that depends on `glog_catkin` (third party), which triggers the download. At the same time, the application `TangoRosStreamer` is a catkin package that depends on `tango_ros_native`. Then, the Android.mk of the `tango_ros_node` (which is part of the `tango_ros_streamer` package) imports `tango_ros_native` module, and finally, the Android.mk file of tango_ros_native includes `miniglog`'s Android.mk (miniglog is not a catkin package).  

After `glog_catkin` triggers the download, but before it finishes, I get the following error message (only the first time, and only if I didn't use -j1):
`make[2]: *** No rule to make target '/home/juan/AndroidStudioProjects/tango_ros_ws/devel/lib/libglog.so', needed by '/home/juan/AndroidStudioProjects/tango_ros_ws/devel/lib/libtango_ros_native.so'.  Stop.`
And that is followed by some errors about tango_ros_native not finding glog/logging.h.
With that being said... I couldn't find a way to fix these dependencies. Any ideas?

PerrineAguiar:
@jubeira FYI, the problem you described with catkin_make already happens on master, so I don't think it is related to the changes of this PR.

smits:
@jubeira I cannot reproduce the glog issue locally, catkin seems to be doing the right thing, downloading and building first and only then it starts building tango_ros_native.  Even without -j1.

smits:
@jubeira For the indigo/opencv2 vs kinetic/opencv3 issue I think we additionally need to build the roscpp_android_ndk with opencv3.  Do you know if this is already the case?

smits:
I get the following error when trying this build:
```
* Where:
Build file '/home/rsmits/tango_ros_ws/src/tango_ros/TangoRosStreamer/tango_ros_node/build.gradle' line: 55

* What went wrong:
A problem occurred evaluating project ':tango_ros_node'.
> /home/rsmits/tango_ros_ws/src/tango_ros/TangoRosStreamer/local.properties (No such file or directory)
```
Could it be that the PR is missing a file?

smits:
Copying the local.properties from RosApp made it work!

smits:
I manually created a local.properties file including:
```
ndk.dir=/opt/android-ndk-r10e
sdk.dir=/opt/android-sdk-linux
```
Now I get an error when building the app:
```
10:48:14.058 [DEBUG] [org.gradle.model.internal.registry.DefaultModelRegistry] Project :app - Registering model element 'tasks.transformClassesWithDexForDebug' (hidden = false)
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] 
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] FAILURE: Build failed with an exception.
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] 
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] * What went wrong:
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] A problem occurred configuring project ':app'.
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] > java.lang.NullPointerException (no error message)
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] 
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] * Try:
10:48:14.068 [ERROR] [org.gradle.BuildExceptionReporter] Run with --stacktrace option to get the stack trace. 
10:48:14.069 [LIFECYCLE] [org.gradle.BuildResultLogger] 
10:48:14.069 [LIFECYCLE] [org.gradle.BuildResultLogger] BUILD FAILED
```
Any ideas on how to resolve this?

smits:
Compiling the project once with Android Studio made it work, not sure how this is related.

adamantivm:
@smits today I've been unsuccessfully trying to run the build on my computer as well - I have way too much garbage that is interfering with the build. I would like to look into that one particular issue you point out. In order to help with that, would you mind clarifying on which version of Ubuntu and ROS are you trying to build? Also, did you do any modification to CMakeLists.txt to allow picking the right version of OpenCV depending on the distribution, like you proposed last time we discussed?

jubeira:
@smits sorry for the delay! Let me try to answer all the points:

- About (my) `catkin_make` failure without -j1, I think there is a race condition that in your case is not manifested. There could be some missing dependency as @adolfo-rt suggests.
- About building roscpp_android with opencv3.. I don't believe that's necessary, I didn't need to do so in my computer (I just changed the dependency to OpenCV 3 in `CMakelists` ).
- The `local.properties` file is required by the `tango_ros_node` module to get the path to the SDK / NDK in your computer in particular. This file is auto generated by Android Studio with the default SDK location, but the true location depends on where each user installs it, so it's fine to exclude this file from the git-tracked list.
- I agree that there is still something to be resolved to make `catkin build` work, but I don't really know how to fix it. As you say, apparently the problem appears only the first time the project is built. I could use some help on this one.

adamantivm:
@smits you won't believe this but I "solved" the `NullPointerException` problem: please check your `local.properties`, specifically if the value of `ndk.dir` doesn't have a trailing space. In my case this was what was causing that build crash and it had originating from copying and pasting from the instructions [here](https://github.com/Intermodalics/tango_ros/blob/master/TangoRosStreamer/README.md), which has that unintended space. Also searching online that error happens due to wrong or missing ndk.dir value in local.properties.

I'm still looking into other issues to make sure catkin build works well, but this particular one Gradle build issue I am positive can be solved this way. Could you please confirm?

adamantivm:
Ok, I can confirm that provided that:
 - #163 is merged into this branch
 - `ndk.dir` in `local.properties` doesn't have a trailing space
 - One is patient with the instructions (they need several updates)

Then building on 16.04 with Kinetic works fine for me. The resulting APK can be found in `devel/share/maven/eu/intermodalics/app/1.0.13/app-1.0.13.apk`

adamantivm:
For the record, these are the main modifications to the new-and-updated README from @juergensturm that I'd suggest to do:

 - Move the `catkin build` part for the end, since it will now also build the Android part (so its pre-requisites need to go first)
 - Include `ros-kinetic-rosjava-build-tools` (or `ros-indigo-rosjava-build-tools`) as an additional dependent package to install
 - Include the instructions to add `local.properties` (and fix the trailing white space!) as it is done [here](https://github.com/Intermodalics/tango_ros/blob/master/TangoRosStreamer/README.md)

With this, I leave this now back your hands @smits

smits:
I can confirm that the path specification of the `ndk.dir` and `sdk.dir` is crucial. In addition I got it building with catkin_tools using `catkin build --no-jobserver`. I guess we need to find out how we can get that `local.properties` correctly in place or rely on `~/.gradle/gradle.properties` or some environment variables instead.

jubeira:
Thanks @smits ! code rebased to master, merging now.
I will open an issue with the summary of this conversation to update the install instructions.

PR #158, opened 2017-02-16
Fix imu timestamp

PerrineAguiar:
This PR fixes the timestamp value of the IMU message.

The error seen in rviz was: 
TF error: [Lookup would require extrapolation into the future. Requested time 1487337410.255000000 but the latest data is at time 1487240508.901808739, when looking up transform from frame [imu] to frame [start_of_service]]


PerrineAguiar:
@smits ptal

PR #159, opened 2017-02-17
Do not publish if transform is empty

PerrineAguiar:
An empty transform is published when drift correction is not enable.
This PR fixes this bug.


PerrineAguiar:
@smits ptal

PR #162, opened 2017-02-23
Improved installation instructions

juergensturm:
This PR refines the installation instructions with more intermediate steps. I believe that I am a good test case for this. 

Unfortunately, there is still a tiny bit missing at the end. With these instructions, everything seems to work (`catkin build` succeeds), but something with Android Studio / gradle is not yet right. Nevertheless, I guess that these instructions are already a better starting point.

If somebody has suggestions, this is the error I'm seeing:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,19 +4,86 @@ It is an Android application for [Tango](https://get.google.com/tango/)-enabled 
 Its main purpose is to provide Tango sensor data to the [ROS](http://wiki.ros.org/) ecosystem in order to easily use the Tango functionalities on robots.  
 You can read the ROS wiki [here](http://wiki.ros.org/tango_ros_streamer).
 
-This work is developed by [Intermodalics](http://www.intermodalics.eu/) in collaboration with [Ekumen](http://www.ekumenlabs.com/).  
+This work is developed by [Intermodalics](http://www.intermodalics.eu/) in collaboration with [Ekumen](http://www.ekumenlabs.com/) and [Tango]().  
 Do not hesitate to give us feedback if something is broken or if you think it lacks some features. The best way to do this is by adding issues to this repository.
 
 # Kickstart
 The app is available in Google's Play Store: https://play.google.com/store/apps/details?id=eu.intermodalics.tango_ros_streamer
 
-# Manual installation
+# Step-by-step installation
+For this guide, we assume that you have a clean installation of Ubuntu 14.04 and we will use ROS Indigo. However, the app should also work on Ubuntu 16.04 in combination with ROS Kinetic.
+
+## Install ROS
+Follow this [guide](http://wiki.ros.org/indigo/Installation/Ubuntu) and go for the Desktop -- Full Install:
+
 ```
-sudo apt-get install python-catkin-tools
+sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
+sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
+sudo apt-get update
+sudo apt-get install ros-indigo-desktop-full
+sudo rosdep init
+rosdep update
+echo "source /opt/ros/indigo/setup.bash" >> ~/.bashrc
+source ~/.bashrc
+```
+
+## Building roscpp_android
+Next, we need to build ```roscpp_android```. A more detailed guide is available [here](http://wiki.ros.org/android_ndk/Tutorials/BuildingNativeROSPackages).
+
+```
+sudo apt-get update
+sudo apt-get install curl git
+curl -sSL https://get.docker.com/ | sudo sh
+mkdir ~/ros-android-ndk
+cd ~/ros-android-ndk
+git clone https://github.com/ekumenlabs/roscpp_android.git
+cd roscpp_android
+./do_docker.sh --portable
+```
+
+Once this has ended, copy the the content of the output ```roscpp_android_ndk``` folder into ```tango_ros/third_party/roscpp_android_ndk/```, except the ```Android.mk``` file and the ```share``` folder.
+
+```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,19 +4,86 @@ It is an Android application for [Tango](https://get.google.com/tango/)-enabled
...
+sudo apt-get install curl git
+curl -sSL https://get.docker.com/ | sudo sh
+mkdir ~/ros-android-ndk
+cd ~/ros-android-ndk
+git clone https://github.com/ekumenlabs/roscpp_android.git
+cd roscpp_android
+./do_docker.sh --portable
+```
+
+Once this has ended, copy the the content of the output ```roscpp_android_ndk``` folder into ```tango_ros/third_party/roscpp_android_ndk/```, except the ```Android.mk``` file and the ```share``` folder.
```
adamantivm: suggestion: maybe make a not about the approximate time this will take? (it's a lot)

README.md
```diff
@@ -4,19 +4,86 @@ It is an Android application for [Tango](https://get.google.com/tango/)-enabled
...
+sudo apt-get install git python-catkin-tools
 mkdir -p ~/tango_ros_ws/src
 cd ~/tango_ros_ws/src
 git clone --recursive git@github.com:Intermodalics/tango_ros.git
 cd ~/tango_ros_ws
 catkin build
 ```  
-To build and run the app: see [here](https://github.com/Intermodalics/tango_ros/blob/master/TangoRosStreamer/README.md).  
+
+## Android Studio
```
adamantivm: maybe reference [this tutorial](http://wiki.ros.org/android/kinetic/Android%20Studio/Download), as you did with the others, which also explains how to install Java, Android Studio and all the dependencies, such that they support rosjava properly?

README.md
```diff
@@ -4,19 +4,86 @@ It is an Android application for [Tango](https://get.google.com/tango/)-enabled
...
+After lauch, open the Android SDK Manager. From SDK Platforms, install Android 4.4 and 5.1 (API level 19 + 22). From SDK Tools, install `NDK` and `Android SDK Build-Tools 21.1.2`. By default, this will be installed to ```$HOME/Android/Sdk```. 
+
+Now extend your PATH variable so that ```ndk-build``` can be executed:
+```
+echo "export PATH=$PATH:$HOME/Android/Sdk/ndk-bundle" >> ~/.bashrc
+source ~/.bashrc
+```
+
+## Building the app
+In Android Studio, choose "Import project" and select the app folder (```~/tango_ros_ws/src/tango_ros/TangoRosStreamer/app```).
```
adamantivm: I think it's better to point to ../TangoRosStreamer  not ../app

README.md
```diff
@@ -4,19 +4,86 @@ It is an Android application for [Tango](https://get.google.com/tango/)-enabled
...
+## Android Studio
+Download from [here](https://developer.android.com/studio/index.html), unzip and run:
+
+```
+cd
+unzip ~/Downloads/android-studio-ide-145.3537739-linux.zip
+./android-studio/bin/studio.sh
+```
+
+After lauch, open the Android SDK Manager. From SDK Platforms, install Android 4.4 and 5.1 (API level 19 + 22). From SDK Tools, install `NDK` and `Android SDK Build-Tools 21.1.2`. By default, this will be installed to ```$HOME/Android/Sdk```.
```
jubeira: I believe this will download Android ndk r13; [TangoRosStreamer Readme](https://github.com/Intermodalics/tango_ros/blob/6558420abbaac4183caa4b46839eb6a833e71e3a/TangoRosStreamer/README.md) uses r10b. Did r13 work for you @juergensturm ? I remember I had problems when I tried to build the app the first time, and I had to switch to r10-b or r10-e.

jubeira:
I think I have a clue about your error @juergensturm . Were you able to generate the apk file? As #152 is still not merged, `catkin build` won't generate it, so you will still need to build it using `gradlew build` or Android Studio's tools. Probably you won't be able to do so as long as this error persists.

Take a look at #163, that should fix it.

README.md
juergensturm: sure, done

README.md
juergensturm: Ok, this tutorial recommends installing Android Studio in /opt and the android-sdk to /opt/android-sdk. Is this really needed and/or should we also recommend that in the README? I would actually prefer to have it in the user's home directory.

README.md
juergensturm: done

README.md
adamantivm: It doesn't make a huge different for your own installation but JFYI the most common / recommended practice is to have them in /opt. I wouldn't publicly recommend something different, in any case I would simply omit it or accept it as OK if it is done differently.

juergensturm:
@jubeira Thanks, this fixed it!

README.md
juergensturm: Ok, I see. Sure, then I'll update it.

juergensturm:
@jubeira @adamantivm PTAL, I addressed your comments.

README.md
```diff
@@ -4,19 +4,105 @@ It is an Android application for [Tango](https://get.google.com/tango/)-enabled
...
 ```  
-To build and run the app: see [here](https://github.com/Intermodalics/tango_ros/blob/master/TangoRosStreamer/README.md).  
+
+## Android Studio
+The steps detailed below are based on this [installation guide](http://wiki.ros.org/android/kinetic/Android%20Studio/Download).
+
+For Android Studio we need Java, so let's install this first:
+```
+sudo apt-get update
+sudo apt-get install openjdk-7-jdk
```
adamantivm: any reason you're using jdk-7? I think jdk-8 is the newest standard (it will soon become required depending on the version of Gradle and Android API used)

juergensturm: It's not available out of the box on Ubuntu 14.04 which is assumed by this tutorial. Let me add a line saying that you should install openjdk-8-jdk on Ubuntu 16.04 to avoid confusion.

juergensturm: I also added this distinction for the ROS installation (indigo vs kinetic).

PR #166, opened 2017-03-02
Refactor/separate nodes from app folder

mcopejans:
@smits PTAL


smits:
@jubeira @adamantivm ptal

adamantivm:
@mcopejans @smits could you please clarify a bit the intent of this move? The folder structure and updated build settings are technically correct, but the resulting structure is a bit unusual. Perhaps if you could clarify what is the objective I could better advice as to whether this is the most recommended approach or another alternative would be preferrable.

mcopejans:
@adamantivm This structure would allow us to reuse the tango_ros_node without copying code.

adamantivm:
@mcopejans what prevents you from doing that in the current structure? In fact, the Tangobot that @jubeira is building is currently re-using the tango_ros_node right now, and it is even documented how to do that: http://wiki.ros.org/tango_ros_streamer/tango_ros_node

mcopejans:
Ok, that would work as well.

I was actually making the folder structure like this to be able to separate the native nodes to another repository one day. There is no specific reason why they should belong to TangoRosStreamer instead of any other app.

adamantivm:
@mcopejans in that case, why not create a new top-level Gradle project altogether for the node only?
This move as is is only moving one module of the project to a different outside folder, but you would still need the entire project to be able to build.

I don't have anything strong against this or against having two separate top-level projects, one for the node and one for the app.

This is some value to having an actual app that uses the node in the same project as the app, for easier testing and documentation on how to use the node. Perhaps my best recommendation would be to create a new top-level Gradle-project-catkin-package, move the node there and add a new very simple application project alongside it which uses the node in the most simple way.

mcopejans:
@adamantivm I wanted to avoid making a new (although simple) application, so I'll go with reusing the node as documented on the wiki: compile 'eu.intermodalics:tango_ros_node:[1.0, 1.1)'.
Thanks for your helpful recommendations!

adamantivm:
For completeness, one good benefit of using the documented method is that you will be using the same integration mechanism that is publicly advertised. So if anything is wrong for the public, you will be able to detect it and fix it, not just for you but also for everyone else.

PR #168, opened 2017-03-03
Fix/unit test unused param

mcopejans:
@smits PTAL


tango_ros_common/tango_ros_native/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(tango_ros_native)
 add_compile_options(-std=c++11)
-set(CMAKE_CXX_FLAGS "-Wunused-parameter -Wunused-variable")
-
```
smits: There is no way around this?

mcopejans: It just prints a massive load of warnings that obscure the actual build output.

PR #172, opened 2017-03-13
Fix: Using OpenCV 2 OR 3

jubeira:
Now that #152 is merged, the OpenCV version was changed to 3. This may cause problems when building the app with Indigo, as the default version is 2 in that distro.
This PR allows to use version 2 or 3, so that the project can be built using Indigo and Kinetic.

This solves #102 .

### DIFF
--- a/tango_ros_common/tango_ros_native/CMakeLists.txt
+++ b/tango_ros_common/tango_ros_native/CMakeLists.txt
@@ -14,6 +14,7 @@ find_package(catkin REQUIRED COMPONENTS
    tf2_ros
    )
 
+# Use OpenCV3 if we are using ROS Kinetic (which is the default for that distro)
 if( $ENV{ROS_DISTRO} STRLESS "kinetic")
   find_package(OpenCV 2 REQUIRED)
 else()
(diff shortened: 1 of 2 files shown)


tango_ros_common/tango_ros_native/CMakeLists.txt
```diff
@@ -14,7 +14,10 @@ find_package(catkin REQUIRED COMPONENTS
    )
 
 find_package(cmake_modules REQUIRED)
-find_package(OpenCV 3 REQUIRED)
+find_package(OpenCV 2 QUIET)
```
PerrineAguiar: I would add a comment to explain why we are doing this.

smits: Why not 
```
if( $ENV{ROS_DISTRO} STRLESS "kinetic")
  find_package(OpenCV 2 REQUIRED)
else()
  find_package(OpenCV 3 REQUIRED)
endif()

jubeira: I will try it out and submit the changes, I think it looks a bit better.
Thanks @smits !

jubeira:
I tested @smits' proposal and it worked just fine in Indigo and Kinetic :+1: 
When I rebased this branch to the current master, I got a compilation error coming from the latest changes in `tango_ros_node.cpp`. I committed a mini fix for that too.
@PerrineAguiar ptal, I think it can be merged.

PR #174, opened 2017-03-14
Fix/tango light color

### LINKED ISSUE #169: Tango connection light intends connection error if successfully connected.
The connection light keeps red / turns red after few seconds, but the app is successfully connected to Tango:
- `rostopic echo tf ` shows device poses
- `rviz` shows fisheye image correctly
- log messages in the app confirm successful connection to Tango (see screenshot below).
screenshot_2017-03-07-14-54-07

PerrineAguiar:
This PR fixes #169.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
@@ -108,12 +108,10 @@ public class RunningActivity extends AppCompatRosActivity implements TangoRosNod
         new DefaultTangoServiceConnection.AfterConnectionCallback() {
             @Override
             public void execute() {
-                Log.w(TAG, "executing DefaultTangoServiceConnection");
                 if (TangoInitializationHelper.isTangoServiceBound()) {
                     if (TangoInitializationHelper.isTangoVersionOk()) {
                         updateTangoStatus(TangoStatus.SERVICE_RUNNING);
                     } else {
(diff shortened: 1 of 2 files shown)


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
```diff
@@ -341,8 +341,10 @@ protected void init(NodeMainExecutor nodeMainExecutor) {
                         != TangoInitializationHelper.ARCH_ERROR) {
             mTangoRosNode = new TangoRosNode();
             mTangoRosNode.attachCallbackListener(this);
+
+            TangoInitializationHelper.checkTangoVersionOk(this);
             TangoInitializationHelper.bindTangoService(this, mTangoServiceConnection);
-            if (TangoInitializationHelper.checkTangoVersionOk(this)) {
+            if (TangoInitializationHelper.isTangoVersionOk()) {
```
mcopejans: I think the implementation of `checkTangoVersionOk` is confusing. It is not clear that it needs to be called for the `mTangoVersionOk` variable to be set. Can you make this a private function, rename to `testTangoVersionOk` and call it inside bindTangoService?

jubeira:
So this happened because sometimes the `ServiceConnection` callback is called before checking the Tango version right? If that's the case, I think @mcopejans' suggestion sounds good; it also makes running the node a bit easier :+1:

PerrineAguiar:
@mcopejans @jubeira Thanks for the review.
I addressed your comment.

PR #177, opened 2017-03-14
Feature/add area learning option

### LINKED ISSUE #131: Allow choice between drift correction and learning mode
Currently only drift correction can be enabled.
Allow the user to choose between drift correction and learning mode (they can not be both enabled).

PerrineAguiar:
This PR adds the option for the user to use learning mode instead of drift correction.

screenshot_20170314-151948

Fixes #131.


adamantivm:
@PerrineAguiar what is the intended workflow for recording and using an ADF? Will "learning mode" here record an ADF? How can the user identify the ADF later? When is the ADF "finished"? How can it be loaded later?
Also, how does the choosing of ADF mode affect which frames are published together with the odometry? And how is this surfaced to library users?
It would be interesting to understand those usage implications.

PerrineAguiar:
@adamantivm This PR is just to add the learning mode options, it does not do anything with the ADF yet. I think it is better to tackle this in follow up PRs.

The workflow for recording and using an ADF and most of your questions will be discussed during tomorrow's meeting, but I can already try to give some ideas.
Yes "learning mode" is recording an ADF. I think one way to handle this ADF would be by having an extra save button in the running activity if learning mode is on. This save button will "finish" the ADF when clicked. The user could give a name to the ADF when saving to be able to identify it later. The loading part could be put in the settings activity since it can not be changed at runtime (tango service needs to disconnect/reconnect with the new tango config).
If drift correction or learning mode is on, or if an ADF is loaded, the additional transform area_description to start_of_service is available.

Maybe I can create a goodle doc to discuss about all of these questions, what do you think?

adamantivm:
Thanks for the clarification @PerrineAguiar . I think a doc about this would be useful for tomorrow's discussion.

PerrineAguiar:
This PR has been slightly modified to follow today's meeting.
screenshot_20170315-171215

For now selecting 'Localization (requires map)' does not have any effect. It will be done in a follow-up PR.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
```diff
@@ -172,20 +172,20 @@ private void callDynamicReconfigure(String paramName, boolean paramValue) {
...
+                    mLog.error("Dynamic Reconfigure failure: " + e.getMessage());
                 }
             });
             mParameterNodeCalledDynamicReconfigure = true;
         } catch (ServiceNotFoundException e) {
-            log.error("Service not found: " + e.getMessage());
+            mLog.error("Service not found: " + e.getMessage());
         } catch (Exception e) {
-            log.error( "Error while calling Dynamic Reconfigure Service: " + e.getMessage());
+            mLog.error( "Error while calling Dynamic Reconfigure Service: " + e.getMessage());
```
jubeira: micro nit: space after (

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -429,8 +429,12 @@ TangoErrorType TangoRosNode::TangoSetupConfig() {
     return result;
   }
 
-  bool enable_drift_correction;
-  node_handle_.param(publisher_config_.enable_drift_correction_param, enable_drift_correction, true);
+  bool enable_drift_correction = false;
+  int localization_mode;
+  node_handle_.param(publisher_config_.localization_mode_param, localization_mode, 1);
```
jubeira: is this "1" the default value? I would use one of the enums for clarity

PR #191, opened 2017-03-20
Feature/record and save map

### LINKED ISSUE #178: ADF integration: Implement recording/saving capability
create / load / save / share ADFs

### LINKED ISSUE #180: ADF Integration: Implement loading capability

### LINKED ISSUE #190: ADF integration: ROS side

PerrineAguiar:
This PR adds the recording/saving/loading map functionalities.

screenshot_20170320-140843

screenshot_20170320-141359

Fixes #178.
Fixes #180.
Fixes #190.


TangoRosStreamer/tango_ros_node/src/main/java/eu/intermodalics/tango_ros_node/TangoRosNode.java
```diff
@@ -64,6 +64,8 @@ public TangoRosNode(String libName) {
     @Override
     public native int shutdown();
 
+    public native boolean saveMap(String mapName);
```
jubeira: I have a doubt about this.
I think it's OK to have a separate method in the node for this purpose, but I'm not sure if this is the "ROS way" of achiving this (e.g. exposing a service that can be called is another option). 
Perhaps exposing a service would make the code more complex than just having a method like this, and I'm not sure if that would have clear benefits in the future. Maybe @adamantivm has some opinion about that.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -428,20 +428,31 @@ TangoErrorType TangoRosNode::TangoSetupConfig() {
...
+    bool enable_drift_correction = false;
+    int localization_mode;
+    node_handle_.param(publisher_config_.localization_mode_param, localization_mode,
+                       (int)LocalizationMode::ODOMETRY);
+    if (localization_mode == LocalizationMode::ONLINE_SLAM) {
+      enable_drift_correction = true;
+    }
+    const char* config_enable_drift_correction = "config_enable_drift_correction";
+    result = TangoConfig_setBool(tango_config_, config_enable_drift_correction, enable_drift_correction);
+    if(result != TANGO_SUCCESS) {
```
jubeira: micro nit: space after "if" (well, this one was here before)

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -698,6 +709,38 @@ void TangoRosNode::StopPublishing() {
   }
 }
 
+bool TangoRosNode::SaveMap(std::string map_name) {
+  TangoErrorType result;
+  TangoUUID map_uuid;
+  result = TangoService_saveAreaDescription(&map_uuid);
+  if(result != TANGO_SUCCESS) {
```
jubeira: same here and elsewhere

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -698,6 +709,38 @@ void TangoRosNode::StopPublishing() {
...
+    LOG(ERROR) << "Error while saving area description, error: " << result;
+    return false;
+  }
+  TangoAreaDescriptionMetadata metadata;
+  result = TangoService_getAreaDescriptionMetadata(map_uuid, &metadata);
+  if(result != TANGO_SUCCESS) {
+    LOG(ERROR) << "Error while trying to access area description metada, error: " << result;
```
jubeira: typo: "matadata", here and elsewhere

TangoRosStreamer/tango_ros_node/src/main/java/eu/intermodalics/tango_ros_node/TangoRosNode.java
PerrineAguiar: Good point. My initial thought was that we would have both: this new API and a ROS service (not yet implemented), but that might be redundant.

smits: +1 on going for the ROS service way and avoiding redundancy

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
smits: @jubeira I guess you suffered from a typo as well :wink: , it should be `metadata` :smile:

jubeira: woops :sweat_smile: yeah, it seems so

PerrineAguiar:
@jubeira @smits Thanks for your review.
I addressed your comments and added the SaveMap service as well as the load functionality. 
As you can see below, the user can choose the map to load in the SettingsActivity via the UI.

screenshot_20170324-112757

To do it via ROS, the user has to set a param containing the UUID of the map to load. It might be nice to add a service which list all available maps (uuid + name). That way it would be easier for the user to set the correct uuid without using the UI.

Some important things to note:
* To be able to generate the Java class of the SaveMap service, genjava needs to be installed with `sudo apt-get install ros-indigo-genjava` <-- need to update installation guide
* It seems that genjava only works with catkin_make (catkin build generates an empty .jar) <-- need to solve this
* When calling the SaveMap service from Java, it returns a connexion error <-- need to solve this

Fixes #180.
Fixes #190.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -340,6 +357,10 @@ TangoRosNode::TangoRosNode() : run_threads_(false) {
   } catch (const image_transport::Exception& e) {
     LOG(ERROR) << "Error while creating image transport publishers" << e.what();
   }
+
+  save_map_service_ = node_handle_.advertiseService<tango_ros_messages::SaveMap::Request,
```
mcopejans: save_map_service_ doesn't seem to be used anywhere else. Can it be removed?

tango_ros_common/tango_ros_messages/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  SaveMap.srv
+)
+
+generate_messages(
+  DEPENDENCIES
+  std_msgs
+)
+
+catkin_package(
+  INCLUDE_DIRS include
```
mcopejans: When compiling this locally in a clean workspace I got an error that no include dir exists in package. So the line above should probably be removed.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
mcopejans: Ok never mind :) You need to keep the service running...

tango_ros_common/tango_ros_messages/CMakeLists.txt
PerrineAguiar: I added the include directory.

mcopejans: I still get errors like this when compiling:
```
make[2]: *** No rule to make target `/home/mcopejans/tangoscan_ws/devel/.private/tango_ros_messages/include/tango_ros_messages/TangoConnectResponse.h', needed by `/home/mcopejans/tangoscan_ws/src/tango_ros/tango_ros_common/tango_ros_messages/include/tango_ros_messages/TangoConnectResponse.h'.  Stop.
```

PerrineAguiar: This error appears when building without the `--no-jobserver` flag.
Just opened #210 to figure out the root cause of the problem.

mcopejans:
@PerrineAguiar Can this be merged?

PerrineAguiar:
I think @smits wanted to have a look before merging.

smits:
I'm currently unable to build this locally due to 
```
Could not find any matches for org.ros.rosjava_messages:tango_ros_messages:[0.0,) as no versions of org.ros.rosjava_messages:tango_ros_messages are available.
     Required by:
         eu.intermodalics:app:1.0.13
```

smits:
Could we avoid having to restart the app when we want to learn a new map, IMO disconnecting and reconnecting to Tango should be sufficient?

PerrineAguiar:
@smits Can we merge this now?

PR #203, opened 2017-03-28
Fix/area description frame in tf tree

### LINKED ISSUE #193: Fix tf transforms in case of Area Learning/COM
AFAIU, when the area learning frame is going to be used (COM or when we load an ADF) we should publish the device pose wrt the area learning frame instead of the Start of Service frame.

In case of VIO:  /start_of_service --> /device
In case of COM/AL: /start_of_service --> /area_learning --> device

As long as we are not localised yet /start_of_service --> /area_learning will be identity.  When we loose localisation we should stick to the old  /start_of_service --> /area_learning until being relocalised instead of starting to publish wrt /start_of_service.

PerrineAguiar:
This PR fixes #193.


tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h
```diff
@@ -174,10 +174,13 @@ class TangoRosNode {
   std::condition_variable color_image_available_;
 
   double time_offset_ = 0.; // Offset between tango time and ros time in s.
+  int localization_mode_;
+  bool first_localized_ = false;
+  bool localization_lost_ = false;
```
jubeira: AFAIU, you are implementing a small state machine right? 
Perhaps I would just declare a variable that represented the status of the localization. For example, an `enum` with three states: starting, localized, localization lost (or more proper names) and take actions according to that state instead of having multiple variables.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -340,6 +340,17 @@ TangoRosNode::TangoRosNode() : run_threads_(false) {
...
     LOG(ERROR) << "Error while creating image transport publishers" << e.what();
   }
+  // Initialize start_of_service_T_area_description as identity.
+  start_of_service_T_area_description_.transform.translation.x = 0.;
+  start_of_service_T_area_description_.transform.translation.y = 0.;
+  start_of_service_T_area_description_.transform.translation.z = 0.;
+  start_of_service_T_area_description_.transform.rotation.x = 0.;
+  start_of_service_T_area_description_.transform.rotation.y = 0.;
+  start_of_service_T_area_description_.transform.rotation.z = 0.;
+  start_of_service_T_area_description_.transform.rotation.w = 1.;
```
jubeira: (minor comment) At least in rosjava, there are some static functions that return a initialized vector / quaternion for some particular cases such as the unit quaternion or the null vector. Perhaps they exist in the Cpp libraries too; it just makes the code a bit shorter.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -495,12 +506,14 @@ TangoErrorType TangoRosNode::TangoSetupConfig() {
...
-
+  TangoCoordinateFramePair pairs[2] = {
+      { TANGO_COORDINATE_FRAME_START_OF_SERVICE,
+        TANGO_COORDINATE_FRAME_DEVICE }, {
+        TANGO_COORDINATE_FRAME_AREA_DESCRIPTION,
+        TANGO_COORDINATE_FRAME_DEVICE }
+  };
   TangoErrorType result;
-  result = TangoService_connectOnPoseAvailable(1, &pair, onPoseAvailableRouter);
+  result = TangoService_connectOnPoseAvailable(2, pairs, onPoseAvailableRouter);
```
jubeira: Well, this one was here before.. perhaps this magic number can be replaced with something more meaningful?

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -610,26 +623,87 @@ void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
...
+            first_localized_ = (area_description_T_device.status_code == TANGO_POSE_VALID);
+            if (first_localized_) {
+              LOG(INFO) << "We are localized!";
+            }
+          }
+        } else if (localization_lost_) {
+          LOG(WARNING) << "Recovering tracking...";
+          // We lost localization, use old start_of_service_T_area_description
+          // to compute area_description_T_device.
+          geometry_msgs::TransformStamped star_of_service_T_device;
```
jubeira: nit: start_of_service_T_device

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -610,26 +623,87 @@ void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
...
+              start_of_service_T_area_description_tf.inverse() * star_of_service_T_device_tf;
+          tf::transformStampedTFToMsg(tf::StampedTransform(
+              area_description_T_device_tf, star_of_service_T_device.header.stamp,
+              toFrameId(TANGO_COORDINATE_FRAME_AREA_DESCRIPTION),
+              toFrameId(TANGO_COORDINATE_FRAME_DEVICE)), area_description_T_device_);
+        }
+        pose_available_.notify_all();
+        pose_available_mutex_.unlock();
+      }
+    } else if (pose->frame.base == TANGO_COORDINATE_FRAME_AREA_DESCRIPTION
```
jubeira: I would add a short comment as in the other cases explaining this one

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -610,26 +623,87 @@ void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
...
+        && pose->frame.target == TANGO_COORDINATE_FRAME_DEVICE
+        && first_localized_) {
+      if (pose->status_code == TANGO_POSE_VALID && pose_available_mutex_.try_lock()) {
```
jubeira: I don't have a really strong opinion about this, but there are two types of checks that are being performed in many places, and perhaps a few helper functions would make the code a little more readable:
First one:
```
if (pose->frame.base == some_frame && pose->frame.target == some_other_frame) {
```
could be written as 
```
if (transformFramesMatch(pose->frame, some_frame, some_other_frame) { // Or some appropriate name
```
And second one:
```
if (pose->status_code == TANGO_POSE_VALID && pose_available_mutex_.try_lock()) {
```
could be replaced by:
```
if (poseCanBePublished(pose)) { // or any proper name
```

Just an opinion though, it's not bad as it is now

PerrineAguiar:
@jubeira Thanks for your review, I addressed your comments.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: I like the first one :+1: 
The second one seems "dangerous", I prefer to keep the mutex lock explicit.

PerrineAguiar:
@smits @mcopejans ptal

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -340,6 +350,13 @@ TangoRosNode::TangoRosNode() : run_threads_(false) {
   } catch (const image_transport::Exception& e) {
     LOG(ERROR) << "Error while creating image transport publishers" << e.what();
   }
+  // Initialize start_of_service_T_area_description as identity.
+  tf::StampedTransform start_of_service_T_area_description =
+      tf::StampedTransform(tf::Transform(tf::Quaternion(0, 0, 0, 1)),
+                           ros::Time::now(),
```
mcopejans: Do you think it make sense to use ros::Time(0) here?

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -607,30 +627,80 @@ void TangoRosNode::PublishStaticTransforms() {
...
+        if (localization_status_ == LocalizationStatus::LOCALIZING) {
+          // Localization is not used or we are not localized yet,
+          // start_of_service_T_area_description is the identity.
+          toTransformStamped(*pose, time_offset_, &area_description_T_device_);
+          area_description_T_device_.header.frame_id =
               toFrameId(TANGO_COORDINATE_FRAME_AREA_DESCRIPTION);
-          area_description_T_start_of_service_.child_frame_id =
+          area_description_T_device_.child_frame_id =
+              toFrameId(TANGO_COORDINATE_FRAME_DEVICE);
+          start_of_service_T_area_description_.header.stamp = ros::Time::now();
```
mcopejans: In case of VIO, the area description frame doesn't have a meaning. Is it possible to only publish area description frame in case of `localization_mode_` equals COM or Area Learning and publish device wrt. SOS in case of VIO?

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -607,30 +627,80 @@ void TangoRosNode::PublishStaticTransforms() {
...
+        } else if (localization_status_ == LocalizationStatus::LOCALIZATION_LOST) {
+          LOG(WARNING) << "Recovering tracking...";
+          // Localization was lost, use old start_of_service_T_area_description
+          // to compute area_description_T_device.
+          geometry_msgs::TransformStamped star_of_service_T_device;
```
mcopejans: Nit. starT_of_service

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -340,6 +350,13 @@ TangoRosNode::TangoRosNode() : run_threads_(false) {
   } catch (const image_transport::Exception& e) {
     LOG(ERROR) << "Error while creating image transport publishers" << e.what();
   }
+  // Initialize start_of_service_T_area_description as identity.
```
mcopejans: Only initialize in case `localization_mode_` equals Area learning or COM.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: I think both would work, but I am not sure which one make more sense.

PerrineAguiar:
@mcopejans Thanks for the review, I addressed your comments.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
mcopejans: Why not publish this frame when it first becomes available? Then this discussion becomes obsolete.

PerrineAguiar: The last commit initializes this frame when it first becomes available.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -590,30 +603,83 @@ void TangoRosNode::PublishStaticTransforms() {
...
+        toTransformStamped(*pose, time_offset_, &start_of_service_T_device_);
+        start_of_service_T_device_.header.frame_id =
+            toFrameId(TANGO_COORDINATE_FRAME_START_OF_SERVICE);
+        start_of_service_T_device_.child_frame_id =
+            toFrameId(TANGO_COORDINATE_FRAME_DEVICE);
+      } else if (localization_status_ == LocalizationStatus::LOCALIZING) {
+        // We are not localized yet.
+        LOG(INFO) << "Not localized yet, please walk around.";
+      } else if (localization_status_ == LocalizationStatus::LOCALIZATION_LOST) {
+        LOG(WARNING) << "Recovering tracking...";
```
mcopejans: This code path is not really necessary in my opinion, but it looks correct :+1:

mcopejans: One more question: Could it be that localization status is set to `LOCALIZATION_LOST` when you were not localized before? This means that adf_T_sos is not available and the multiplication will be wrong... In this case I would just remove this code path.

PerrineAguiar: It can happen if the adf_T_device pose is invalid and we still did not get any valid sos_T_adf pose. Is it something that could actually happen in Tango?

mcopejans: Maybe for a short time... I'm not sure.

Suggestion to solve this: Either remove this code path or add a bool `sos_T_adf_initialized_` which initializes to false whenever you connect to Tango.

PerrineAguiar: I went for the second option :+1:

smits:
After the discussion in #193 we can conclude the initial implementation of the tf tree was correct after all.

Issue #187, opened 2017-03-17
State for "connecting"


adamantivm:
It is very hard to tell from the green / red status light whether it is trying to connect to tango or if it already failed - it is in both cases still red.

It would be nice to have an intermediate "connecting" state (yellow? spinning wheel?)

PerrineAguiar:
+1 for an intermediate state.
The Tango light always red bug was fixed in #174.

PR #207, opened 2017-03-30
Feature/tango ros connect disconnect service

### LINKED ISSUE #208: Add Tango connect/disconnect service
Necessary to convert Tango ROS node to a nodelet in the future.

mcopejans:
Adds service to connect and disconnect from Tango. Closes #208
@PerrineAguiar PTAL


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
```diff
@@ -159,6 +162,24 @@ public void onTangoRosErrorHook(int returnCode) {
...
+        int count = 0;
+        while (count < 20 && !mTangoServiceClientNode.callTangoConnectService(TangoConnectRequest.CONNECT)) {
+            try {
+                count++;
+                Log.e(TAG, "Trying to connect to Tango, attempt " + count);
+                Thread.sleep(200);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        }
```
mcopejans: @PerrineAguiar Do you know a way to know when the `execute` hook from a node is finished?

jubeira:
@mcopejans I have a suggestion: you could add some rosjava standard listeners when you execute the node instead of overriding `onStart`. And you can also perform some action `onShutdown`. Something like this:
```java
mNodeMainExecutor.execute(mTangoRosNode, nodeConfiguration,
                new ArrayList<NodeListener>(){{
                    add(new DefaultNodeListener() {
                        @Override
                        public void onStart(ConnectedNode connectedNode) {
                            // Do something onStart
                        }

                        @Override
                        public void onShutdown() {
                            // do something on shutdown
                        }
                    });
                }});
```
With this you won't need to add more code to the node, and it's a standard rosjava mechanism. We tried something alike in Tangobot and it works fine!

mcopejans:
@jubeira Thanks, this solution is already a lot cleaner than what I did! However, it doesn't solve everything yet since onStart is already called when execute did not finish completely.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
```diff
@@ -190,7 +198,7 @@ private void switchTangoLight(final TangoStatus status) {
         runOnUiThread(new Runnable() {
             @Override
             public void run() {
-                if (status == TangoStatus.SERVICE_RUNNING) {
+                if (status == TangoStatus.SERVICE_RUNNING || status == TangoStatus.SERVICE_CONNECTED) {
```
PerrineAguiar: Don't we need `&&` instead?

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
```diff
@@ -283,6 +305,30 @@ public void run() {
...
+
+    @Override
+    public void onTangoDisconnectServiceFinish(int response, String message) {
+        if (response != TangoConnectResponse.TANGO_SUCCESS) {
+            Log.e(TAG, "Error disconnecting from Tango: " + response + ", message: " + message);
+            switchTangoLight(TangoStatus.SERVICE_NOT_CONNECTED);
+            displayToastMessage(R.string.tango_disconnect_error);
+            return;
+        }
+
```
mcopejans: I added an extra topic `/tango/status` which is published from `tango_ros_node` when (dis)connection failed or succeeded. This is required to make the Tango Status lights work when the `tango/connect` service is called directly from `ROS`.

mcopejans:
@PerrineAguiar @jubeira Thanks for your input and review, I addressed all comments.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
mcopejans: The status cannot be two different things at the same time?

PerrineAguiar: True...then maybe turn the light green only if `status == TangoStatus.SERVICE_CONNECTED`?

mcopejans: ... and orange in case Tango Service is running?

PerrineAguiar: Good point.
Just realized that this is related to #187, so it can be solved later when tackling this issue.
Sorry for the noise.

PR #209, opened 2017-03-31
Feature/remove publish data booleans

### LINKED ISSUE #60: Consider removing publish_* parameters.
Maybe we should rely on the number of subscribers, if no-one is subscribed we just don't publish, otherwise we do.

PerrineAguiar:
This PR removes the navigation drawer from the app and the `publish_*` parameters from dynamic reconfigure. Instead, data is published if at least one subscriber exists.

Fixes #60.


PerrineAguiar:
@jubeira I did not find an easy way to know how many subscribers to a certain transform exist. I guess it is because /tf is not a typical topic. It internally keeps the tf tree so you might need some transform even if no one seems to subscribe to it from the outside.
In the application case you mentioned, you could use only the poses you are interested in, even if some others are published on /tf. Or this is not a solution?

PerrineAguiar:
@jubeira Thanks for the review.
@smits @mcopejans ptal

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
```diff
@@ -77,118 +69,16 @@ public ParameterNode(Activity activity, String[] dynamicParamNames, String[] par
...
-                        }
-                    }
-                    editor.commit();
-                }
-            }
-        });
-
         // Set ROS params according to preferences.
         for (String paramName : mParamNames) {
             String stringValue = mSharedPreferences.getString(paramName, "1");
```
mcopejans: Unrelated to this PR, but why is the default value "1"?

TangoRosStreamer/app/src/main/res/layout/running_activity.xml
```diff
@@ -86,19 +86,4 @@
...
-        android:layout_gravity="right"
-        android:background="#282828">
-
-        <FrameLayout
-            android:id="@+id/preferencesFrame"
-            android:layout_height="match_parent"
-            android:layout_width="match_parent">
-        </FrameLayout>
-    </LinearLayout>
 </android.support.v4.widget.DrawerLayout>
```
mcopejans: Remove drawerlayout completely.

mcopejans:
Perhaps [TopicManager](http://docs.ros.org/api/roscpp/html/classros_1_1TopicManager.html) could help to count the number of subscribers on a tf topic? Never used it myself, so I'm not sure if it works for `/tf`.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/ParameterNode.java
PerrineAguiar: This is solved in #191.

PerrineAguiar:
It seems that `ros::TopicManager` is used internally by ros but is not aimed to be used for our purpose.

adamantivm:
One question: what if you're want to somehow configure the node to *not* publish a given topic, no matter if a subscriber is present or not? Iiuc this is now not possible, right? I think It would be very nice to be able to force given services off.

jubeira:
Yes @PerrineAguiar, that is a solution, but as @adamantivm points out, it would be nice to have a way of stopping the threads to save processing time.

PR #226, opened 2017-04-13
Add map name and uuid in save map service response

### LINKED ISSUE #223: SaveMap service: reply with UUID and map name in separate fields
Currently it is reported in the reply message in a message, and hence is cumbersome to get the UUID (and name) to load the map automatically.

PerrineAguiar:
This PR adds the map name and uuid to the SaveMap service response.

Fixes #223.


dvanthienen:
LGTM!

PR #230, opened 2017-04-14
Fix condition variable usage and add logging

smits:
These were some fixes to make sure the mesh publishing doesn't deadlock, I also added some additional logging when something unexpected happens when generating the mesh


PerrineAguiar:
@smits Thanks.
It is ok if I resolve the conflicts myself and merge it afterwards?

PR #234, opened 2017-04-14
Fix nodelet namespace

### LINKED ISSUE #232: Fix topic/param/service namespace

PerrineAguiar:
This PR fixes the nodelet namespace so that all topics/params/services start with the same prefix: /tango

Fixes #232.


mcopejans:
FYI @malteim

PR #235, opened 2017-04-14
Fix tango permission at first run

### LINKED ISSUE #231: Fix permission pop up on first run

PerrineAguiar:
This PR fixes 3 issues related to tango permision and first run:
* Fix app crash in SettingsActivity at first run
* For the first run, ask for Tango permissions only when going back to the RunningActivity
* If a Tango permission is not granted, display a toast message instead of finishing the app

Fixes #231.


PerrineAguiar:
@smits Thanks for the review.

PR #236, opened 2017-04-18
Add missing plugin for image raw

PerrineAguiar:
This PR adds missing plugin for raw images.

### DIFF
--- a/TangoRosStreamer/nodelet_manager/src/main/jni/jni.cc
+++ b/TangoRosStreamer/nodelet_manager/src/main/jni/jni.cc
@@ -16,7 +16,9 @@
 #include <jni.h>
 
 #include <glog/logging.h>
+#include <image_transport/publisher_plugin.h>
 #include <nodelet/loader.h>
+#include <pluginlib/class_loader.h>
 #include <ros/ros.h>
 
 #include "tango_helper.h"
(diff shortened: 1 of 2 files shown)


mcopejans:
LGTM. Is it possible to add a warning when some plugins cannot be found?

PerrineAguiar:
@mcopejans Thanks for your review. As you suggested I added a warning when a plugin is not available. I am not sure if it would have prevented this bug but it is still a good thing.

mcopejans:
Throwing an exception would certainly have prevented this bug...

TangoRosStreamer/nodelet_manager/src/main/jni/jni.cc
```diff
@@ -68,6 +70,18 @@ Java_eu_intermodalics_nodelet_1manager_NodeletManager_execute(
     }
     LOG(INFO) << "Finished loading nodelets.";
 
+    // Check that all necessary plugins are available.
+    pluginlib::ClassLoader<image_transport::PublisherPlugin> image_transport_pub_loader("image_transport", "image_transport::PublisherPlugin");
+    if (!image_transport_pub_loader.isClassAvailable("image_transport/raw_pub")) {
+      LOG(WARNING) << "Plugin image_transport/raw_pub is not available.";
+      return 1;
+    } else if (!image_transport_pub_loader.isClassAvailable("image_transport/compressed_pub")) {
```
mcopejans: This should be just `if`, not `else if`.

PR #239, opened 2017-04-19
Fix/no maps to choose

### LINKED ISSUE #225: No maps show up if you go to the settings page too fast after starting the app.
If you go to the settings page after starting the app, before Tango is connected, no maps show up, even after the connection is established.
You have to go back to the main menu and back into settings to see the maps.

PerrineAguiar:
With this PR the Choose Map preference in the SettingsActivity gets updated when connecting to Tango.

Fixes #225.


PerrineAguiar:
@mcopejans ptal

PerrineAguiar:
@mcopejans It does not work for the first run. In this case we start the app with the settings activity and connect to Tango only when going to the running activity.
I will make an issue for it to keep track of this.

PR #250, opened 2017-04-24
Restart tango after saving map from UI button

### LINKED ISSUE #249: New saved map not shown in list
After saving a new map via the UI button and going in the settings, the new map is not in the available maps list.

PerrineAguiar:
Fixes #249.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
@@ -305,6 +305,8 @@ public class RunningActivity extends AppCompatRosActivity implements NodeletMana
                 }
             });
             displayToastMessage(R.string.save_map_success);
+            // Restart Tango to be able to load the new map.
+            mTangoServiceClientNode.callTangoConnectService(TangoConnectRequest.RECONNECT);
         } else {
             Log.e(TAG, "Error while saving map: " + message);
             displayToastMessage(R.string.save_map_error);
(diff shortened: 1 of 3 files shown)


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/RunningActivity.java
```diff
@@ -305,6 +305,8 @@ public void run() {
                 }
             });
             displayToastMessage(R.string.save_map_success);
+            // Restart Tango to be able to load the new map.
+            mTangoServiceClientNode.callTangoConnectService(TangoConnectRequest.RECONNECT);
```
smits: Does this come with a notification?

PerrineAguiar: A toast message is displayed when saving the map ended successfully, and another one when reconnecting to Tango is finished.
Maybe we can add the information that we are reconnecting Tango in the first message, e.g.: "Map saved successfully. Tango is restarting now."?

smits: When it restarts, will it switch to localisation mode in the saved map?

PerrineAguiar: No, we restart Tango to restart motion tracking (saving an ADF stops it) and update the list of maps available. Instead of restarting Tango which can be confusing, we could only update the list of maps and add a toast message saying that motion tracking has stopped.

PerrineAguiar:
@smits As discussed during the meeting, instead of restarting Tango, we only disconnect it and inform the user in a toast message.

PR #254, opened 2017-04-26
Feature/preapre new release

PerrineAguiar:
PR preparing the new release of the app.


adamantivm:
@PerrineAguiar one thing, and sorry I missed this in previous PR's, note that if you change the name of the tango_ros_node module, you will change the name with which the library package and artifact is released, leaving behind the older version with no further updates in the repo. Is that the intended behavior? Did you consider using the same package and artifact name with the new implementation?

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/AboutActivity.java
```diff
@@ -48,7 +48,7 @@ protected void onCreate(Bundle savedInstanceState) {
         TextView textViewVersion = (TextView) findViewById(R.id.text_version);
         try {
             PackageInfo packageInfo = getPackageManager().getPackageInfo(getPackageName(), 0);
-            textViewVersion.setText(packageInfo.versionName + "-" + packageInfo.versionCode);
+            textViewVersion.setText(packageInfo.versionName);
```
jubeira: This doesn't show the patch version, is that OK?

PerrineAguiar: It was actually an error, we were using the version code as the patch version, which is wrong. The patch version should be included in the version name and is totally independent from the version code.

PerrineAguiar:
@adamantivm @jubeira Thanks for your comments.
I agree that changing the library name is not ideal but as @jubeira said, the new name reflects better what it is. To have the new functionalities, updating the module name and the class is necessary. Do you think it is a big issue? I am asking you because you are the only one I know using the tango_ros_node.
I will update the wiki entry as soon as this is merged.

adamantivm:
I think it's not a big issue, especially since we're the only ones using it
(that we know) so far, so it's a good time to change before it has more
uses.

PerrineAguiar:
@smits ptal

Issue #219, opened 2017-04-11
YS device: App is seg faulting when connecting to Tango


PerrineAguiar:
The seg fault seems to appear only on YS device:
* when running the on device test
* when disconnecting/reconnecting using the new tango connect service
```
04-11 12:08:57.556 11319-11535/? I/tango_client_api: TangoErrorType TangoService_connect(void*, TangoConfig): Successfully connected to Tango.
04-11 12:08:57.584 170-11734/? I/Camera3-OutputStream: enqueue first Preview frame or first video frame
04-11 12:08:57.586 9000-11669/? I/tango: gpu-freak-descriptor.cc:305 Rotation invariance for ocv FREAK descriptor is: 1
04-11 12:08:57.586 9000-11669/? I/tango: gpu-freak-descriptor.cc:307 No. of octaves to be used in opencv descriptor module 4
04-11 12:08:57.657 162-162/? I/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
04-11 12:08:57.657 162-162/? I/DEBUG: Build fingerprint: 'google/tango/yellowstone:4.4.2/KOT49H.161129/161129001:userdebug/dev-keys'
04-11 12:08:57.657 162-162/? I/DEBUG: Revision: '0'
04-11 12:08:57.658 162-162/? I/DEBUG: pid: 11319, tid: 11345, name: Binder_3  >>> eu.intermodalics.tango_ros_streamer <<<
04-11 12:08:57.658 162-162/? I/DEBUG: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0000005c
04-11 12:08:57.672 9000-11669/? I/tango: gpu-freak-descriptor.cc:287 Loaded LuT for CUDA FREAK.
04-11 12:08:57.675 9000-11669/? I/tango: gpu-freak-descriptor.cc:387 Input image is from the fisheye camera.
04-11 12:08:57.675 9000-11669/? I/tango: gpu-freak-descriptor.cc:421 Extracting Lens Invariant FREAK descriptors using camera intrinsics.
04-11 12:08:57.703 170-11735/? I/Camera3-OutputStream: enqueue first Preview frame or first video frame
04-11 12:08:57.766 162-162/? I/DEBUG:     r0 0000005c  r1 4b570b38  r2 4dbca000  r3 0000005c
[...]

PR #256, opened 2017-04-28
Fix/error message when timout tango connection

PerrineAguiar:
This PR adds a missing error message when the number of tries to call the tango connect service exceed a certain number.


PerrineAguiar:
@smits I tested the new apk on Lenovo devices, I think we can release it once this PR is merged.

However, on Yellowstone devices the app crashes when reconnecting to Tango. This was already reported in #219, but we decided to focus on Lenovo devices. One solution would be to release the new apk only for Lenovo?

TangoRosStreamer/app/build.gradle
```diff
@@ -27,7 +27,7 @@ dependencies {
     compile 'org.ros.rosjava_messages:sensor_msgs:[0.3,)'
     compile 'org.ros.rosjava_messages:geometry_msgs:[0.3,)'
     compile 'org.ros.rosjava_messages:dynamic_reconfigure:[0.3,)'
-    compile 'org.ros.rosjava_messages:tango_ros_messages:[0.0,)'
+    compile 'org.ros.rosjava_messages:tango_ros_messages:1.0.0'
```
jubeira: Suggestion: use [1.0.0, 1.1) format to support minor changes that don't break the API just in case.

PerrineAguiar:
@smits did you take a look? If this is lgtm I can make the new release.

PR #263, opened 2017-05-05
Fix/permission and getmaps ys

PerrineAguiar:
This PR fixes two issues found while testing on YS:
* start the tango nodelet only after asking for tango permissions
* returning an empty string if there is no map on the device yet.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
@@ -122,6 +122,8 @@ public class RunningActivity extends AppCompatRosActivity implements TangoNodele
     private boolean mMapSaved = false;
     private HashMap<String, String> mUuidsNamesHashMap;
     private BroadcastReceiver mRestartTangoAlertReceiver;
+    private boolean mAdfPermissionHasBeenAsked = false;
+    private boolean mDatasetPermissionHasBeenAsked = false;
 
     // UI objects.
     private TextView mUriTextView;
@@ -495,7 +497,13 @@ public class RunningActivity extends AppCompatRosActivity implements TangoNodele
                 // No Tango permissions granted by the user.
                 displayToastMessage(R.string.tango_permission_denied);
             }
-            if (requestCode == REQUEST_CODE_DATASET_PERMISSION) {
+            if (requestCode == REQUEST_CODE_ADF_PERMISSION) {
+                mAdfPermissionHasBeenAsked = true;
+            }
+            if (requestCode ==  REQUEST_CODE_DATASET_PERMISSION) {
+                mDatasetPermissionHasBeenAsked = true;
+            }
+            if (mAdfPermissionHasBeenAsked && mDatasetPermissionHasBeenAsked) {
                 Log.i(TAG, "initAndStartRosJavaNode");
                 initAndStartRosJavaNode();
             }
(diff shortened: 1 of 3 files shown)


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -384,11 +385,11 @@ private void saveUuidsNamestoHashMap() {
         mTangoServiceClientNode.callGetMapUuidsService();
     }
 
-    private void getTangoPermission(String permissionType) {
+    private void getTangoPermission(String permissionType, int requestCode) {
         Intent intent = new Intent();
         intent.setAction(REQUEST_TANGO_PERMISSION_ACTION);
         intent.putExtra(EXTRA_KEY_PERMISSIONTYPE, permissionType);
-        startActivityForResult(intent, REQUEST_CODE_TANGO_PERMISSION);
+        startActivityForResult(intent, requestCode);
```
mcopejans: Following current Tango documentation ([here](https://developers.google.com/tango/apis/java/java-user-permissions)), this can be done in a simpler way:
```
startActivityForResult(
    Tango.getRequestPermissionIntent(Tango.PERMISSIONTYPE_ADF_LOAD_SAVE),
    Tango.TANGO_INTENT_ACTIVITYCODE);
```

PerrineAguiar: This is when using the Java API. In our case we use the C API so I used [this](https://developers.google.com/tango/apis/c/c-user-permissions).

mcopejans: Ok! Looks good.

mcopejans: The current [activity result callback](https://github.com/Intermodalics/tango_ros/blob/d2571aff8b6d69baf60e830488ee44a27a0d1354/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java#L498) only reacts on positive dataset permission. Is a similar case also required for ADF permissions?

mcopejans: The current [code](https://github.com/Intermodalics/tango_ros/blob/d2571aff8b6d69baf60e830488ee44a27a0d1354/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java#L498) only instantiates Java node in case dataset permissions are granted. Is the same required for ADF permissions?

PerrineAguiar:
@mcopejans Thanks for the reiview, I addressed your comments.

mcopejans:
OK, the implementation is quite conservative because permissions for both adf and dataset are always required even if not needed. This could be addressed as a follow up?
LGTM

PerrineAguiar:
@mcopejans Both permissions are not required to be allowed, but to be asked, i.e. we wait for the user to answer both permissions pop up before starting the node.
I will add some comments in the code so things are clearer.

mcopejans:
Asking would also not be required in all cases. Only if specific settings are enabled.

PerrineAguiar:
Yes, I'll make a follow up to address this.

PR #265, opened 2017-05-05
Allow to disable depth camera + move parameter node to tango_common

mcopejans:
Allows to enable/disable the depth camera using Settings Activity or the ROS parameter server.

The PR also moves the parameter node to the tango_ros_common module since it is written in a generic way and can be reused in other Tango-ROS apps.

### DIFF
--- a/TangoRosStreamer/app/src/main/res/values/strings.xml
+++ b/TangoRosStreamer/app/src/main/res/values/strings.xml
@@ -75,7 +75,7 @@
     <string name="pref_create_new_map_summary">Enable to create a new localization map. This map can be used for localization later.</string>
 
     <string name="pref_enable_depth">Enable depth camera</string>
-    <string name="pref_enable_depth_key">enable_depth_camera</string>
+    <string name="pref_enable_depth_key">enable_depth</string>
     <string name="pref_enable_depth_summary">Enable depth camera when connecting to Tango.</string>
 
     <string name="pref_localization_mode_key">localization_mode</string>
--- a/TangoRosStreamer/app/src/main/res/values/strings.xml
+++ b/TangoRosStreamer/app/src/main/res/values/strings.xml
@@ -74,6 +74,10 @@
     <string name="pref_create_new_map_key">create_new_map</string>
     <string name="pref_create_new_map_summary">Enable to create a new localization map. This map can be used for localization later.</string>
 
+    <string name="pref_enable_depth">Enable depth camera</string>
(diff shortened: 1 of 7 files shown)


mcopejans:
Perhaps we can do the same for color images? I'm guessing this could also save some battery power.

adamantivm:
+1

TangoRosStreamer/app/src/main/res/values/strings.xml
```diff
@@ -74,6 +74,10 @@
     <string name="pref_create_new_map_key">create_new_map</string>
     <string name="pref_create_new_map_summary">Enable to create a new localization map. This map can be used for localization later.</string>
 
+    <string name="pref_enable_depth">Enable depth camera</string>
+    <string name="pref_enable_depth_key">enable_depth_camera</string>
```
PerrineAguiar: The key should be the same as the parameter name, i.e. `enable_depth`.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -351,7 +351,9 @@ TangoErrorType TangoRosNode::TangoSetupConfig() {
     return result;
   }
   const char* config_enable_depth = "config_enable_depth";
-  result = TangoConfig_setBool(tango_config_, config_enable_depth, true);
+  bool enable_depth;
+  node_handle_.param<bool>(ENABLE_DEPTH, enable_depth, true);
+  result = TangoConfig_setBool(tango_config_, config_enable_depth, enable_depth);
```
PerrineAguiar: Initialize the ros parameter in the `onInit` function.

PerrineAguiar:
@mcopejans I will make a new issue to do the same for color image.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
mcopejans: Does this mean that if the parameter is changed between ROS startup and (re)connection to Tango, it would not be picked up? Is this what we want?

PerrineAguiar: No, of course these lines should stay to pick up changes when reconnecting to Tango. What I meant is that the ROS param should be initialized in the `onInit` function, like we do for the others.
Sorry for the confusion, I added this comment to the wrong line.

mcopejans:
Currently implemented like other settings, but needs fix described here: https://github.com/Intermodalics/tango_ros/issues/268

PR #267, opened 2017-05-08
[NOT FOR MERGE] Add log messages to investigate pose not published on tf issue

PerrineAguiar:
[NOT FOR MERGE]

This PR brings some changes to make investigation on #259 easier:
* adding some extra log messages
* changing the code to publish pose data on separate topics next to /tf


smits:
I would rather not merge this in master as the logging would get spammed a lot, could we make an apk from this branch and distribute that one instead?

PerrineAguiar:
@smits Just rolled out an apk with these changes in the alpha release.

PR #271, opened 2017-05-09
Switch preference has bugs in Android 4.4

PerrineAguiar:
This PR replaces the SwitchPreference which enables the depth camera by a checkbox. The reason is that having multiple SwitchPreferences has a bug in Android 4.4 (Tango Development Kit device), i.e. when switching one preference the other one also switches.

Source: https://issuetracker.google.com/issues/36941388

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/CustomSwitchPreference.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/CustomSwitchPreference.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Intermodalics All Rights Reserved.
+ * Copyright 2017 Intermodalics All Rights Reserved.
  *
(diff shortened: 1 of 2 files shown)


mcopejans:
Wow, that's a funny bug!

Changes here are LGTM. For consistency, we could alter the SwitchPreference for pref_create_new_map as well or implement the workaround as described in the issue?

PerrineAguiar:
@mcopejans You are right, I implemented the workaround instead.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/CustomSwitchPreference.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2016 Intermodalics All Rights Reserved.
```
mcopejans: Nit. 2017

PerrineAguiar:

PR #272, opened 2017-05-09
Ask for permission only when needed

### LINKED ISSUE #270: Ask for Tango permissions only when needed instead of at start

PerrineAguiar:
Fixes #270.


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -380,6 +376,12 @@ public void onTangoStatus(int status) {
             saveUuidsNamestoHashMap();
             mParameterNode.setPreferencesFromParameterServer();
             mMapSaved = false;
+            if (mParameterNode.getBoolParam(getString(R.string.pref_create_new_map_key)) ||
+                    mParameterNode.getIntParam(getString(R.string.pref_localization_mode_key)) == 3)
+                getTangoPermission(EXTRA_VALUE_ADF, REQUEST_CODE_ADF_PERMISSION);
+            if (!mParameterNode.getStringParam(getString(R.string.dataset_datasets_path_key)).isEmpty() &&
+                    !mParameterNode.getStringParam(getString(R.string.dataset_uuid_key)).isEmpty())
+                getTangoPermission(EXTRA_VALUE_DATASET, REQUEST_CODE_DATASET_PERMISSION);
```
mcopejans: Isn't it required to get Tango permissions before connecting to Tango (`SERVICE_CONNECTED`)?

PerrineAguiar: The documentation is not really clear about it but I think not, they are required only for specific Tango API calls like: `TangoService_getAreaDescriptionUUIDList`, `TangoService_getAreaDescriptionMetadata`, `TangoService_saveAreaDescription` or `TangoService_Experimental_getDatasetUUIDs`.

What we could also do is, each time we do one of this call from the C++ code, somehow trigger this intent in the Java code. I think it could be possible using a ros topic.

mcopejans: Tango Connect returns TANGO_NO_DATASET_PERMISSION or similar if permissions are not granted BEFORE connecting.

See: https://github.com/googlesamples/tango-examples-c/blob/master/tango_client_api/include/tango_client_api.h#L76

PerrineAguiar:
@mcopejans Thanks for your comment. I changed the code to trigger the permission intent from c++ using the tango status topic, ptal.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -295,6 +297,20 @@ TangoErrorType TangoRosNode::OnTangoServiceConnected() {
   return TANGO_SUCCESS;
 }
 
+void TangoRosNode::RequestADFPermissionAndWaitForAnswer() {
+  UpdateAndPublishTangoStatus(TangoStatus::NEED_TO_REQUEST_ADF_PERMISSION);
+  while (tango_status_ != TangoStatus::ADF_PERMISSION_REQUEST_ANSWERED) {
+    ros::Duration(0.1).sleep();
+  }
```
mcopejans: Is it possible to implement this using a service call? Sleeping does not look like an elegant solution. Why did you choose not to request permissions before connecting on the Java side.

mcopejans (changes requested):

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: I chose to trigger permission request from the C++ side because this is where we know exactly when and which permission we need. For example, if the user reconnect to Tango using the ros command  line tool, the Java side does not know that a service call to `/tango/connect has` been made.
Here we are using the `/tango/status` topic to trigger the permission request from C++ to Java, and know when the request has been answered.

smits: Is abusing the /tango/status the best way to do this? Why not using a ros service call between the java side (server) and the c++ side (client)?

PerrineAguiar:
@mcopejans @smits Thanks for your comment.
I implemented a service to request permissions as suggested, ptal.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -291,6 +293,34 @@ TangoErrorType TangoRosNode::OnTangoServiceConnected() {
...
+}
+
+void TangoRosNode::RequestDatasetPermission() {
+  tango_ros_messages::RequestPermission srv;
+  srv.request.permission = tango_ros_messages::RequestPermission::Request::DATASET_PERMISSION;
+  if (request_permission_service_.call(srv)) {
+    if (srv.response.granted) {
+      LOG(INFO) << "Dataset permission has been requested and granted by the user.";
+    } else {
+      LOG(WARNING) << "Dataset permission has been requested but not granted by the user.";
```
mcopejans: If I understand correctly, a permission that was not granted will lead to a connection failure? I would resend the request in this case or force the user to go back to the settings screen. The app cannot continue in any other way.

mcopejans (approved):
The code looks good, but I added one comment to improve the app flow. For this feature, I think it is useful to have a schema in mind that illustrates what happens when each permission is granted vs. not granted.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: I think it will indeed fail to connect if we try to load/record dataset while dataset premission is not granted. I would like to test it but I did not find a proper way to revoke the dataset permission.

PerrineAguiar:
@mcopejans Thanks for the review.

I think the schema for ADF and dataset permission is as follow:
* ADF permission is needed only when saving an ADF or accessing its metadata. In our case it means that if the user does not grant ADF permission, he will not be able to save a map or see the list of available maps.
* Dataset permission is needed when loading/recording dataset. If the user does not grant dataset permission while loading or recording a dataset, connection to Tango will fail.

mcopejans:
Jenkins failed with the following error, is this expected:
```
ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h:43:50: fatal error: tango_ros_messages/RequestPermission.h: No such file or directory
 #include <tango_ros_messages/RequestPermission.h>
```

mcopejans:
test this please

PerrineAguiar:
@mcopejans Yes, I think one line is missing [here](https://github.com/Intermodalics/tango_ros/pull/272/files#diff-6f8bf9417a4843254a65eed5d30a1ad1R43).
It should be:
```
RequestPermission
RequestPermissionRequest
RequestPermissionResponse
```

mcopejans:
Yay, one point for Jenkins :)

meyerj:
retest this please

PR #275, opened 2017-05-10
Feature/button to retry ros connection

### LINKED ISSUE #194: Use MasterChecker instead of doing master check in TangoRosNode

### LINKED ISSUE #242: Button to retry connect to ROS
See https://github.com/Intermodalics/TangoApps/pull/298.

PerrineAguiar:
With this PR, if the connection to the ROS master failed, we show some animations and a snackbar with a reconnect button.
It also uses the `MasterConnectionChecker` (from [tangobot](https://github.com/ekumenlabs/tangobot/blob/master/tangobot_app/app/src/main/java/com/ekumen/tangobot/application/MasterConnectionChecker.java)) to check the connection to the ROS master before starting the nodes.

Fixes #242.
Fixes #194.


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -168,22 +176,14 @@ public void execute() {
         }
     );
 
-    /**
-     * Implements TangoRosNode.CallbackListener.
-     */
-    public void onNodeletManagerError(int returnCode) {
```
mcopejans: Can this function also be removed from nodelet manager?

TangoRosStreamer/app/src/main/res/animator/master_uri_text_animation.xml
```diff
@@ -0,0 +1,14 @@
...
+        android:valueFrom="#676767"
+        android:valueTo="#FF0000"
+        android:interpolator="@android:anim/linear_interpolator"
+        android:valueType="colorType"
+        android:repeatCount="3"
+        android:repeatMode="reverse"
+        android:duration="400"
+        />
+
+</set>
```
mcopejans: Nit newline.

TangoRosStreamer/app/src/main/res/values/strings.xml
```diff
@@ -24,35 +25,20 @@
 
     <string name="save_map_dialog_title">Enter the name of the map</string>
     <string name="uuids_names_map">uuids_names_map</string>
+    <string name="ros_status">ros_status</string>
 
     <!-- Strings related to error messages -->
 
     <string name="network_error">No network found! Please check your network configuration and restart the application.</string>
-    <string name="ros_init_error">Unable to initialize ROS node. Do you have a roscore running?</string>
+    <string name="ros_init_error">Unable to connect to ROS master. Do you have a roscore running?</string>
```
mcopejans: An other reason for connection failure could be that ROS master URI is incorrect. Do you agree to remove the latter part `Do you have a roscore running?` I think it could be confusing.

tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/MasterConnectionChecker.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        @Override
+        protected void onPostExecute(Boolean result) {
+            if (result && mUserHook != null) {
+                mUserHook.onSuccess(mOnSuccessPayload);
+            }
+        }
+    }
+}
```
mcopejans: Nit. newline

TangoRosStreamer/app/src/main/res/values/strings.xml
```diff
@@ -24,35 +25,20 @@
...
-
-    <string name="publish_laser_scan">Laser scan</string>
-    <string name="publish_laser_scan_key">publish_laser_scan</string>
-
-    <string name="publish_fisheye_camera">Fisheye image</string>
-    <string name="publish_fisheye_camera_key">publish_fisheye_camera</string>
-
-    <string name="publish_color_camera">Color image</string>
-    <string name="publish_color_camera_key">publish_color_camera</string>
+    <string name="tango_permission_denied">Be aware that you will not be able to use all functionalities from TangoRosStreamer</string>
```
mcopejans: Is this text still up to date after the permission PR went in? This might be more correct: `Connection to Tango will fail if requested permissions are not granted.`

PerrineAguiar: The connection to Tango will not fail but the user will not be able to save a map for example.

PerrineAguiar:
@mcopejans Thanks for your review, I addressed your comments.

TangoRosStreamer/app/src/main/res/values/strings.xml
mcopejans: The connection can fail with "NO_DATASET_PERMISSIONS" or similar for ADF if you try to load a dataset or ADF at connection  time.

PerrineAguiar: I tested to load an ADF after denying the permission and it is working fine. 
From the tango_client_api.h:
_Returns @c TANGO_NO_DATASET_PERMISSION if the
///     config_enable_dataset_recording flag was enabled, but the user has not
///     given permissions to read and write datasets.
TangoErrorType TangoService_connect(void* context, TangoConfig config);_

I suppose it is similar for ADF, so it can fail at connection time when recording ADF/dataset.

However, if the user is not recording a map, ADF permission is still needed to get the list of maps available. In this case Tango connect will not fail, but the user will not be able to see the list of maps in the SettingsActivity.

Therefore the toast message could be something like:
`Permission denied: Connection to Tango can fail and some functionnalities will not work properly.`

mcopejans: According to docs, it is indeed not similar for ADF. Only for datasets. New text looks good except for typo `functionalities`

PerrineAguiar:

PR #276, opened 2017-05-16
Changes tf_broadcast to tf_static_broadcast .

malteim:


tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -823,10 +823,10 @@ void TangoRosNode::PublishDevicePose() {
       std::unique_lock<std::mutex> lock(device_pose_thread_.data_available_mutex);
       device_pose_thread_.data_available.wait(lock);
       if (publish_pose_on_tf_) {
-        tf_broadcaster_.sendTransform(start_of_service_T_device_);
+        tf_static_broadcaster_.sendTransform(start_of_service_T_device_);
```
mcopejans: This is a local workaround and should not go in master ATM.

PR #279, opened 2017-05-16
Bugfix/gradle version

dvanthienen:


TangoRosStreamer/build.gradle
```diff
@@ -5,7 +5,9 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.2'
+        classpath('com.android.tools.build:gradle:2.3.2') {
+            force = true
```
smits: Why is the force needed?

smits:
Can you explain the issue you are trying to solve?

PR #282, opened 2017-05-19
Adds enable color setting and fixes sync between settings activity and ROS parameter server

### LINKED ISSUE #266: Add ros param to enable fisheye/color image

### LINKED ISSUE #268: Fix consistency settings Settings Activity vs. TangoRosNode
Current situation: If a setting does not exist on ROS parameter server, it will be initialized on the ROS parameter server using its default value. These settings are synced to the Settings Activity AFTER successful Tango connection.

This can lead to inconsistencies since the default value of a setting can be different from the one that was saved in shared preferences from the previous run of the app.

Suggested solution:
1. Sync all settings from Settings Activity to ROS parameter server BEFORE Tango connection. 
Advantage: Default values of settings only need to be known in Settings Activity and are not duplicated in Tango ROS node anymore.
2. Before quitting the app and when starting the Settings Activity, all parameters from ROS parameter server need to be synced to Settings activity.

mcopejans:
Closes #266 
Adds setting to enable or disable the color camera to save battery. Only works on API > 23

Closes #268 
Fixes sync between settings activity and ROS parameter server.
- When parameter node is started, all settings are downloaded from ROS parameter server and local settings are uploaded to parameter server.
- Before starting the settings activity, parameters are downloaded from parameter server.
- After quitting the settings activity, parameters are uploaded to parameter server.
- Before quitting RunningActivity, parameter are downloaded from parameter server.
- Settings are downloaded before restarting Tango.
- When calling the `/tango/connect` service directly, the C++ ROS node takes care of getting parameters from the server. These settings will be synced with the local settings whenever the SettingsActivity is started next time (or when app quits).

Issues:
- At the first run of Settings activity or when ROS connection failed, the local settings are overwritten by the parameter server when parameter node starts.
- ROS parameters cannot be changed if settings activity is open. ROS parameters will be overwritten whenever a user returns to the RunningActivity.

Is this behavior correct? Are the issues acceptable?

### DIFF
--- a/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h
+++ b/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h
@@ -78,6 +78,7 @@ const std::string DATASET_PATH_PARAM_NAME = "dataset_datasets_path";
 const std::string DATASET_UUID_PARAM_NAME = "dataset_uuid";
 const std::string USE_FLOOR_PLAN_PARAM_NAME = "use_floor_plan";
 const std::string ENABLE_DEPTH = "enable_depth";
+const std::string ENABLE_COLOR_CAMERA = "enable_color";
 const std::string PUBLISH_POSE_ON_TF_PARAM_NAME = "publish_pose_on_tf";
 const std::string PUBLISH_POSE_ON_TOPIC_PARAM_NAME = "publish_pose_on_topic";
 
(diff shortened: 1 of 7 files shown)


tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h
```diff
@@ -78,6 +78,7 @@ const std::string DATASET_PATH_PARAM_NAME = "dataset_datasets_path";
 const std::string DATASET_UUID_PARAM_NAME = "dataset_uuid";
 const std::string USE_FLOOR_PLAN_PARAM_NAME = "use_floor_plan";
 const std::string ENABLE_DEPTH = "enable_depth";
+const std::string ENABLE_COLOR_CAMERA = "enable_color";
```
PerrineAguiar: The param name and the setting key should be the same.

mcopejans:
@PerrineAguiar Thanks for your review. Addressed all comments.

PR #285, opened 2017-05-22
Fix publish pose on TF

mcopejans:
Publish pose on tf parameters are not duplicated in SettingsActivity, so they have to be synced from C++. (Bug introduced in PR #282)


mcopejans:
@PerrineAguiar Thanks!

PR #287, opened 2017-05-26
Fix/enable cameras settings

PerrineAguiar:
This PR fixes some bugs regarding the enable depth/color camera parameters:
* When starting with a fresh roscore, they were set to false instead of true by default.
* When reconnecting to Tango to change their value, it was hanging when trying to stop the threads.

It also disables the switch preference for the color camera in case the android device is running with API < 23.

### DIFF
--- a/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
+++ b/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
@@ -771,43 +771,43 @@ void TangoRosNode::StopPublishing() {
   if (run_threads_) {
     run_threads_ = false;
     if (device_pose_thread_.publish_thread.joinable()) {
-      if (!tango_data_available_ || !publish_pose_on_tf_
-          || !publish_pose_on_topic_) {
+      if (!tango_data_available_ || !publish_pose_on_tf_ ||
+          !publish_pose_on_topic_) {
         device_pose_thread_.data_available.notify_all();
       }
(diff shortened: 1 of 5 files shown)


PerrineAguiar:
@mcopejans Not for review yet, I still need to make the changes discussed during today's meeting.

PerrineAguiar:
@mcopejans Ready for review.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -771,43 +771,43 @@ void TangoRosNode::StopPublishing() {
   if (run_threads_) {
     run_threads_ = false;
     if (device_pose_thread_.publish_thread.joinable()) {
-      if (!tango_data_available_ || !publish_pose_on_tf_
-          || !publish_pose_on_topic_) {
+      if (!tango_data_available_ || !publish_pose_on_tf_ ||
+          !publish_pose_on_topic_) {
```
mcopejans: I forgot in which cases it is necessary to call notify_all before joining.

PerrineAguiar: In this case it is because we call [wait](http://en.cppreference.com/w/cpp/thread/condition_variable/wait), which block the publisher thread until `notify_all()` is called.
Normally `notify_all()` is called in the Tango callback but in some conditions the call to `notify_all()` is not reached.

PR #298, opened 2017-05-30
Advertise depth and color topics only if camera enable

### LINKED ISSUE #294: Do not advertise camera topics if they are not enable.

PerrineAguiar:
Fixes #294.


tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -166,12 +166,6 @@ void TangoRosNode::onInit() {
   area_description_T_start_of_service_publisher_ =
       node_handle_.advertise<geometry_msgs::TransformStamped>(
           AREA_DESCRIPTION_T_START_OF_SERVICE_TOPIC_NAME, queue_size, latching);
-  point_cloud_publisher_ =
```
mcopejans: What is the advantage of moving the publishers to the connect method? Did you measure noticeable performance gain by moving the publisher initialization to tango connect? One disadvantage is that publishers need to be reinitialized when doing a disconnect connect cycle.

PerrineAguiar: (Including @meyerj since he initially reported the issue)

@smits With this PR the parameter setting has influence on the advertisement each time we connect to Tango.

The reason behind this change was not to gain performance but to reduce risks of user mistakes. If a topic is listed in `rostopic list` and the camera is actually disable it can be confusing for the user.

meyerj: Correct. If a stream-like topic is advertised I would expect to see data on it. At least it simplifies reasoning on what a node is doing from the output of `rostopic list` or `rosnode info`.

An alternative would be to activate the image streams on demand once a subscriber connects, if that is possible without interruptions for other streams.

smits: I follow the reasoning that if a topic is advertised data can be expected. +1

PR #302, opened 2017-05-31
Fix shared preferences null

### LINKED ISSUE #301: Shared preferences null
Shared preferences can be null in ParameterNode if ROS did not connect successfully. This leads to a NullPointerException if a user tries to go to the SettingsActivity to e.g. adapt Master URI.

mcopejans:
SharedPreferences object in ParameterNode can be null if ROS master not start successfully. This leads to a crash due to NullPointerException if a user tries to go to SettingsActivity in this case to e.g. set the correct Master URI.

Closes #301


mcopejans:
@PerrineAguiar PTAL

PR #310, opened 2017-06-06
Make optional callback in tango service node and add explicit method to set listener

### LINKED ISSUE #308: Misleading signature in TangoServiceClientNode
The [constructor of the TangoServiceClientNode](https://github.com/Intermodalics/tango_ros/blob/master/tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/TangoServiceClientNode.java#L71) class requires an Activity but immediately casts it to another class `TangoServiceClientNode.CallbackListener`. This is bad coding practice since it leads developers to build programs that will fail at runtime with cryptic problems, causing long debug times.

This is particularly serious since the documentation explaining how to integrate `tango_ros_node` in other applications is grossly outdated at the moment.

I suggest making the callback optional and moving it to a separate, explicit setListener call.

PerrineAguiar:
Fixes #308.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
@@ -77,7 +77,7 @@ import tango_ros_messages.TangoConnectRequest;
 import tango_ros_messages.TangoConnectResponse;
 
 public class RunningActivity extends AppCompatRosActivity implements
-        SaveMapDialog.CallbackListener, TangoServiceClientNode.TangoServiceCallbackListener {
+        SaveMapDialog.CallbackListener, TangoServiceClientNode.CallbackListener {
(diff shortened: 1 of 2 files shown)


PerrineAguiar:
@adamantivm ptal

adamantivm:
Thank you for this @PerrineAguiar . I think it is good as is. 

Only a couple of optional suggestions:
 - I think since this is an inner interface inside a class that is already called TangoServiceSomething, it is not necessary to add the TangoService prefix to the interface name, imho.
 - One alternative way to cover for the case where no listener is set is to create a DefaultCallbackListener class that doesn't do anything (or that logs a message indicating no callback for each method), and set a new instance of a DefaultCallbackListener by default. This has the added benefit that if someone wants to implement only a given listener, they can extend the DefaultCallbackListener and only override the functions that the user is interested in overriding - vs. implementing an interface which forces the user to implement all methods.

PerrineAguiar:
@adamantivm Thanks for your comment, I followed your suggestions.

PR #312, opened 2017-06-08
Fix/ros vs tango timestamp

### LINKED ISSUE #299: Timestamp issue between tango and ros

PerrineAguiar:
This fixes #299.


PerrineAguiar:
FYI @adamantivm @jubeira

mcopejans:
@PerrineAguiar Code changes are LGTM. Do you intend to merge this or should we keep it in a feature branch to avoid having forwards-backwards changes in master?

Some questions that could help tracking down the actual timestamp issue: Do you have an idea why the timestamp is wrong? How big is the offset? Is the time offset due to accuracy or due to an error in the calculation? Is the offset always constant?

PerrineAguiar:
@mcopejans Thanks for the review.
I would merge this PR (but keep the issue opened) as this is blocking Ekumen at the moment.

The reason of the issue is explained in #299. The problem comes from the calculation of the offset and the fact that we do not have a way to convert Tango time to system time. 
The error of the offset seems to be between 0.09 to 0.18 s (values taken from tf exceptions when trying to use the Tango ROS Streamer with the navigation stack). It varies between runs and also during the same run.

PerrineAguiar:
@jubeira @adamantivm @mcopejans 
I am now using CLOCK_BOOTTIME as suggested by Ian. This allows us to calculate accurately the time offset between Tango and ROS. I tested it with the turtlebot and the navigation stack and it seems to resolve the tf issue.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -272,7 +272,10 @@ TangoErrorType TangoRosNode::OnTangoServiceConnected() {
     LOG(ERROR) << "Error, could not get a first valid pose.";
     return TANGO_INVALID;
   }
-  time_offset_ =  ros::Time::now().toSec() - pose.timestamp;
+  struct timespec res_boot;
+  clock_gettime(CLOCK_BOOTTIME, &res_boot);
+  time_offset_ = ros::Time::now().toSec() -
+      (res_boot.tv_sec + (double) res_boot.tv_nsec / 1e9);
```
mcopejans: Perhaps move to separate function for cleanliness and testability?

PR #315, opened 2017-06-16
Add missing returns for GetMap service functions

smits:
Small fix, we should return something, I'm not entirely sure whether we should do this is just return true always.

### DIFF
--- a/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
+++ b/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
@@ -1118,6 +1118,7 @@ bool TangoRosNode::GetMapName(
     const tango_ros_messages::GetMapName::Request &req,
     tango_ros_messages::GetMapName::Response &res) {
   res.map_name = GetMapNameFromUuid(req.map_uuid);
+  return res.map_name.empty();
 }
 
 bool TangoRosNode::GetMapUuids(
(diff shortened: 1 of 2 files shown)


tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -1118,6 +1118,7 @@ bool TangoRosNode::GetMapName(
     const tango_ros_messages::GetMapName::Request &req,
     tango_ros_messages::GetMapName::Response &res) {
   res.map_name = GetMapNameFromUuid(req.map_uuid);
+  return res.map_name.empty();
```
mcopejans: Is this return value used anywhere? From the function name `GetMapName`, I would expect that it returns the map name instead of a boolean. Suggestion: rename to `GetMapNameServiceRequestCallback`.

smits:
Based on http://wiki.ros.org/roscpp/Overview/Services#Class_Methods we should always return true IMHO.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
smits: +1, that would indeed make it clear according to http://wiki.ros.org/roscpp/Overview/Services#Class_Methods we should probably always return true.

lounick:
If it always guarantees to succeed yes. In a ros service you can return false if the result could not be computed for some reason. In the call of GetMapNameFromUuid you may fail to get the string.

Actually according to TangoAreaDescriptionMetadata_get the value of the char* can be NULL if the key is not found or not set. I don't you are allowed to create a string instance with a NULL value.

A better signature could be bool GetMapNameFromUuid(const std::string& map_uuid, std::string &map_name) and return false on any of the errors.

Sorry for the hijack :D

smits:
@lounick Thanks for the Hijack, I realised the same in the meantime and implemented a refactor accordingly. PTAL @lounick and @mcopejans

lounick:
@smits LGTM

PR #320, opened 2017-06-19
Fix load newly saved map

PerrineAguiar:
This fixes the feature to load  a newly saved map.
It got broken when working on syncing app preferences and ROS params.


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -354,6 +354,7 @@ public void onClick(View view) {
                             editor.putString(getString(R.string.pref_localization_mode_key), "3");
                             editor.putString(getString(R.string.pref_localization_map_uuid_key), mapUuid);
                             editor.commit();
+                            mParameterNode.uploadPreferencesToParameterServer();
```
mcopejans: Nice bug fix!
Suggestion: By making a helper function in ParameterNode, you can encapsulate the settings. This way, only the parameter node manages all settings and if e.g. localization mode enum "3" would change, this is a logical place to fix this. Example code:
```
class ParameterNode {
  void changeSettingsToFinishMappingAndStartLocalizingInNewMap() {
    (...) // Add above functionality to this helper function.
  }
}
```

PerrineAguiar:
@mcopejans Thanks for your review and your suggestion.

PR #325, opened 2017-06-21
Catch parameter class cast exception instead of crashing

### LINKED ISSUE #321: App crashes if rosparam value does not match its type
Crashes with:
```
06-20 14:40:14.865 8390-8422/? E/AndroidRuntime: FATAL EXCEPTION: pool-1-thread-7
                                                 Process: eu.intermodalics.tango_ros_streamer, PID: 8390
                                                 org.ros.exception.ParameterClassCastException: Cannot cast parameter to: java.lang.Boolean
                                                     at org.ros.internal.node.parameter.DefaultParameterTree.get(DefaultParameterTree.java:206)
                                                     at org.ros.internal.node.parameter.DefaultParameterTree.getBoolean(DefaultParameterTree.java:225)
                                                     at org.ros.internal.node.parameter.DefaultParameterTree.getBoolean(DefaultParameterTree.java:230)
                                                     at eu.intermodalics.tango_ros_common.ParameterNode.setPreferencesFromParameterServer(ParameterNode.java:111)
                                                     at eu.intermodalics.tango_ros_common.ParameterNode.syncLocalPreferencesWithParameterServer(ParameterNode.java:69)
                                                     at eu.intermodalics.tango_ros_common.ParameterNode.onStart(ParameterNode.java:61)
                                                     at org.ros.internal.node.DefaultNode$5.run(DefaultNode.java:508)
                                                     at org.ros.internal.node.DefaultNode$5.run(DefaultNode.java:505)
                                                     at org.ros.concurrent.EventDispatcher.loop(EventDispatcher.java:43)
                                                     at org.ros.concurrent.CancellableLoop.run(CancellableLoop.java:56)
                                                     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
                                                     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
                                                     at java.lang.Thread.run(Thread.java:818)
                                                  Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Boolean
                                                     at java.lang.Class.cast(Class.java:1505)
```

Should not crash but use default value and print error instead.

PerrineAguiar:
Fixes #321.

### DIFF
--- a/tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/ParameterNode.java
+++ b/tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/ParameterNode.java
@@ -122,7 +122,16 @@ public class ParameterNode extends AbstractNodeMain implements NodeMain {
                         editor.putString(paramName, stringValue);
                     }
                 } catch (ParameterClassCastException e) {
-                    mLog.error("Preference " + paramName + " can not be set from parameter server. " + e.getMessage());
+                    if (mParamNames.get(paramName) == "boolean") {
+                        try {
+                            Integer intValue = mConnectedNode.getParameterTree().getInteger(NodeNamespaceHelper.BuildTangoRosNodeNamespaceName(paramName), 0);
+                            editor.putBoolean(paramName, !intValue.equals(0));
+                        } catch (ParameterClassCastException e2) {
+                            mLog.error("Preference " + paramName + " can not be set from parameter server.", e2);
+                        }
+                    } else {
+                        mLog.error("Preference " + paramName + " can not be set from parameter server.", e);
+                    }
                 }
             }
         }
--- a/tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/ParameterNode.java
+++ b/tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/ParameterNode.java
@@ -21,6 +21,7 @@ import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
 
 import org.apache.commons.logging.Log;
+import org.ros.exception.ParameterClassCastException;
 import org.ros.namespace.GraphName;
 import org.ros.node.AbstractNodeMain;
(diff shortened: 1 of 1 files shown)


tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/ParameterNode.java
```diff
@@ -107,17 +108,21 @@ public void setPreferencesFromParameterServer() {
...
+                    if (mParamNames.get(paramName) == "int_as_string") {
+                        Integer intValue = mConnectedNode.getParameterTree().getInteger(NodeNamespaceHelper.BuildTangoRosNodeNamespaceName(paramName), 0);
+                        editor.putString(paramName, intValue.toString());
+                    }
+                    if (mParamNames.get(paramName) == "string") {
+                        String stringValue = mConnectedNode.getParameterTree().getString(NodeNamespaceHelper.BuildTangoRosNodeNamespaceName(paramName), "");
+                        editor.putString(paramName, stringValue);
+                    }
+                } catch (ParameterClassCastException e) {
+                    mLog.error("Preference " + paramName + " can not be set from parameter server. " + e.getMessage());
```
mcopejans: Catching an exception is a good thing, but in case of boolean you can add additional logic to try to cast it manually if value equals 0 or 1.

adamantivm: Another small note (sorry to hijack the comment): I always recommend try to pass the exception as a second parameter to log.error methods, instead of appending the exception message string. That will give more and more precise information to the user debugging the error.

adamantivm: In other words: `mLog.error("Preference " + paramName + " can not be set from parameter server.", e);`

PerrineAguiar:
@mcopejans @adamantivm Thanks for reviewing, I addressed your comments.

PR #331, opened 2017-06-26
Feature/load save navigation map

### LINKED ISSUE #289: Save/load 2D navigation map
See http://wiki.ros.org/map_server

PerrineAguiar:
This PR adds two new features:
* Save a navigation map from the occupancy grid data as a pgm (image) and yaml (metadata) files. To do so the save_map service has been changed so that the user can choose which map to save (localization map, navigation map or both).
* Load a navigation map from a pgm and yaml file and publish it as an occupancy grid. To do so a new service has been added.

EDIT:
The UI of the app has been updated : 
* The existing "Save Map" button saves both the localization and the navigation map
* A new button "Load Navigation Map" has been added. For now it just requires the name of the navigation map to load.

Fixes #289.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
@@ -342,11 +342,8 @@ public class RunningActivity extends AppCompatRosActivity implements
     }
 
     public void onClickOkSaveMapDialog(final String mapName) {
-        if (mapName == null || mapName.isEmpty()) {
-            Log.e(TAG, "Map name is null or empty, unable to save the map");
-            displayToastMessage(R.string.name_error);
-            return;
-        }
+        assert(mapName !=null && !mapName.isEmpty());
+        mSaveMapButton.setEnabled(false);
         new AsyncTask<Void, Void, Void>() {
             @Override
             protected Void doInBackground(Void... params) {
@@ -354,7 +351,6 @@ public class RunningActivity extends AppCompatRosActivity implements
                 return null;
             }
         }.execute();
-        mSaveMapButton.setEnabled(false);
     }
 
     @Override
@@ -388,11 +384,8 @@ public class RunningActivity extends AppCompatRosActivity implements
     }
 
     public void onClickOkLoadOccupancyGridDialog(final String occupancyGridName) {
-        if (occupancyGridName == null || occupancyGridName.isEmpty()) {
-            Log.e(TAG, "Name is null or empty, unable to load the occupancy grid");
-            displayToastMessage(R.string.name_error);
-            return;
-        }
+        assert(occupancyGridName !=null && !occupancyGridName.isEmpty());
+        mLoadOccupancyGridButton.setEnabled(false);
         new AsyncTask<Void, Void, Void>() {
             @Override
             protected Void doInBackground(Void... params) {
@@ -400,7 +393,6 @@ public class RunningActivity extends AppCompatRosActivity implements
                 return null;
             }
         }.execute();
-        mLoadOccupancyGridButton.setEnabled(false);
     }
 
     @Override
@@ -660,7 +652,6 @@ public class RunningActivity extends AppCompatRosActivity implements
 
     private void restartTango() {
         if (mParameterNode != null) mParameterNode.setPreferencesFromParameterServer();
-        updateLoadAndSaveMapButtons();
         mTangoServiceClientNode.callTangoConnectService(TangoConnectRequest.RECONNECT);
     }
 
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
@@ -388,14 +388,6 @@ public class RunningActivity extends AppCompatRosActivity implements
                             mParameterNode.changeSettingsToLocalizeInMap(mapUuid, getString(R.string.pref_create_new_map_key),
                                     getString(R.string.pref_localization_mode_key), getString(R.string.pref_localization_map_uuid_key));
                             restartTango();
-                            mLoadOccupancyGridButton.setEnabled(false);
-                            new AsyncTask<Void, Void, Void>() {
-                                @Override
-                                protected Void doInBackground(Void... params) {
-                                    mTangoServiceClientNode.callLoadOccupancyGridService(mapName);
-                                    return null;
-                                }
-                            }.execute();
                         }
                     });
                     mSnackbarLoadNewMap.show();
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
@@ -46,6 +46,7 @@ import android.widget.ImageView;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import org.apache.commons.io.FilenameUtils;
 import org.ros.address.InetAddressFactory;
 import org.ros.android.NodeMainExecutorService;
 import org.ros.android.NodeMainExecutorServiceListener;
@@ -56,6 +57,7 @@ import org.ros.node.NodeConfiguration;
 import org.ros.node.NodeListener;
 import org.ros.node.NodeMainExecutor;
 
+import java.io.File;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -131,6 +133,7 @@ public class RunningActivity extends AppCompatRosActivity implements
     private boolean mAdfPermissionHasBeenAnswered = false;
     // True after the user answered the dataset permission popup (the permission has not been necessarily granted).
     private boolean mDatasetPermissionHasBeenAnswered = false;
+    private ArrayList<String> mOccupancyGridNameList = new ArrayList<String>();
 
     // UI objects.
     private Menu mToolbarMenu;
@@ -296,11 +299,13 @@ public class RunningActivity extends AppCompatRosActivity implements
         saveMapDialog.show(manager, "SaveMapDialog");
     }
 
-    private void showLoadOccupancyGridDialog(boolean firstTry) {
+    private void showLoadOccupancyGridDialog(boolean firstTry, java.util.ArrayList<java.lang.String> nameList) {
         FragmentManager manager = getFragmentManager();
         LoadOccupancyGridDialog loadOccupancyGridDialog = new LoadOccupancyGridDialog();
         Bundle bundle = new Bundle();
+        bundle.putBoolean(getString(R.string.show_load_occupancy_grid_empty_key), nameList.isEmpty());
         bundle.putBoolean(getString(R.string.show_load_occupancy_grid_error_key), !firstTry);
+        bundle.putStringArrayList(getString(R.string.list_names_occupancy_grid_key), nameList);
         loadOccupancyGridDialog.setArguments(bundle);
         loadOccupancyGridDialog.setStyle(DialogFragment.STYLE_NORMAL, R.style.CustomDialog);
         loadOccupancyGridDialog.show(manager, "LoadOccupancyGridDialog");
@@ -335,7 +340,17 @@ public class RunningActivity extends AppCompatRosActivity implements
         mLoadOccupancyGridButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
-                showLoadOccupancyGridDialog(true);
+                mOccupancyGridNameList = new ArrayList<String>();
+                File occupancyGridDirectory = new File("/sdcard/tango_ros_streamer/occupancy_grids");
+                if (occupancyGridDirectory != null && occupancyGridDirectory.isDirectory()) {
+                    File[] files = occupancyGridDirectory.listFiles();
+                    for (File file : files) {
+                        if (FilenameUtils.getExtension(file.getName()).equals("yaml")) {
+                            mOccupancyGridNameList.add(FilenameUtils.removeExtension(file.getName()));
+                        }
+                    }
+                }
+                showLoadOccupancyGridDialog(true, mOccupancyGridNameList);
             }
         });
         updateLoadAndSaveMapButtons();
@@ -383,7 +398,7 @@ public class RunningActivity extends AppCompatRosActivity implements
         }
     }
 
-    public void onClickOkLoadOccupancyGridDialog(final String occupancyGridName) {
+    public void onClickItemLoadOccupancyGridDialog(final String occupancyGridName) {
         assert(occupancyGridName !=null && !occupancyGridName.isEmpty());
         mLoadOccupancyGridButton.setEnabled(false);
         new AsyncTask<Void, Void, Void>() {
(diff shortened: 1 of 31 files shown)


adamantivm:
@PerrineAguiar are you going to make the new LoadMap service call available through the `TangoServiceClientNode`?

PerrineAguiar:
@adamantivm Yes, we will also update the UI to be able to load a navigation map.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -364,6 +383,32 @@ public void onClick(View view) {
         }
     }
 
+    public void onClickOkLoadNavMapDialog(final String mapName) {
+        if (mapName == null || mapName.isEmpty()) {
+            Log.e(TAG, "Map name is null or empty, unable to load the navigation map");
+            displayToastMessage(R.string.map_name_error);
+            return;
```
mcopejans: This error handling is not very user friendly. Is it a valid use case to load a map with an empty map name? Or to press Ok without a map being selected?

- Try to implement the dialogFragment in such a way that the map name is  never empty by e.g. not listing the maps with empty map name.
- Disable the Ok button until a valid map was selected. Otherwise, only allow cancel.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -364,6 +383,32 @@ public void onClick(View view) {
...
+        }
+        new AsyncTask<Void, Void, Void>() {
+            @Override
+            protected Void doInBackground(Void... params) {
+                mTangoServiceClientNode.callLoadNavMapService(mapName);
+                return null;
+            }
+        }.execute();
```
mcopejans: Make sure that the user cannot load another map while the service is already running. Disable the load map button in this case or open a progress dialog to keep the user waiting.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -364,6 +383,32 @@ public void onClick(View view) {
...
+    }
+
+    @Override
+    public void onLoadNavMapServiceCallFinish(boolean success, final String message) {
+        if (success) {
+            displayToastMessage(R.string.load_nav_map_success);
+            Log.i(TAG, message);
+        } else {
+            Log.e(TAG, "Error while loading navigation map: " + message);
+            displayToastMessage(R.string.load_nav_map_error);
```
mcopejans: A toast message is only displayed on the screen for a very short time. It is therefore easy to miss. Does it make sense to show the load map dialog again when the map failed to load with extra message in red text, e.g. Map "map name" failed to load for some reason, try again with a different map<

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -603,7 +648,7 @@ public void onError(Throwable t) {
 
     private void restartTango() {
         if (mParameterNode != null) mParameterNode.setPreferencesFromParameterServer();
-        updateSaveMapButton();
+        updateLoadAndSaveMapButtons();
```
mcopejans: It looks like this function needs to be called in many places in this activity to keep the UI in sync with the Tango status. Shouldn't it be sufficient to only add it to `onTangoStatus`?

If Tango is restarted, the status should go from connected to disconnected and then to connected again, so the function onTangoStatus will be called in this cycle.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/LoadNavMapDialog.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright 2016 Intermodalics All Rights Reserved.
```
mcopejans: Nit. 2017

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/LoadNavMapDialog.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public View onCreateView(LayoutInflater inflator, ViewGroup container,
+                             Bundle savedInstanceState) {
+        View dialogView = inflator.inflate(R.layout.dialog_load_nav_map, null);
+        getDialog().setTitle("Enter name of navigation map to load");
+        mNameEditText = (EditText) dialogView.findViewById(R.id.map_name);
+        dialogView.findViewById(R.id.load_nav_map_ok).setOnClickListener(this);
+        dialogView.findViewById(R.id.load_nav_map_cancel).setOnClickListener(this);
+        setCancelable(false);
```
mcopejans: Why is the dialog set to `not cancelable`? Also in L58, the callback function for the cancel button is defined so this is contradictory.

Consider the following use case: A user might have accidentally clicked the load map button without actually needing to load a map. In this case, he would like to cancel the dialog and without being forced to load a navigation map and reconnect to Tango.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/LoadNavMapDialog.java
```diff
@@ -0,0 +1,78 @@
...
+    public void onAttach(Context context) {
+        super.onAttach(context);
+        mCallbackListener = (CallbackListener) context;
+    }
+
+    @Override
+    public View onCreateView(LayoutInflater inflator, ViewGroup container,
+                             Bundle savedInstanceState) {
+        View dialogView = inflator.inflate(R.layout.dialog_load_nav_map, null);
+        getDialog().setTitle("Enter name of navigation map to load");
```
mcopejans: Add to string resources.

TangoRosStreamer/app/src/main/res/layout/dialog_load_nav_map.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Intermodalics All Rights Reserved. -->
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
mcopejans: Consider using an `AlertDialog` and autocreate a list with existing map names on the system.

An AlertDialog can show a title, up to three buttons and a list of selectable items (or a custom layout if you would still like to do something special).

tango_ros_common/tango_ros_common/src/main/java/eu/intermodalics/tango_ros_common/TangoServiceClientNode.java
```diff
@@ -183,12 +191,13 @@ private Boolean saveMap(String mapName) throws ServiceNotFoundException {
                         SaveMap._TYPE);
 
         SaveMapRequest saveMapRequest = mConnectedNode.getServiceRequestMessageFactory().newFromType(SaveMap._TYPE);
+        saveMapRequest.setRequest((byte)0); // Save loc and nav maps.
```
mcopejans: Use the enum from the message definition here.

tango_ros_common/tango_ros_messages/srv/SaveMap.srv
```diff
@@ -1,6 +1,11 @@
+int8 LOC_AND_NAV_MAPS=0
+int8 ONLY_LOC_MAP=1
```
mcopejans: It would make sense to encode this binary:
LOAD_LOCALIZATION_MAP = 1
LOAD_NAVIGATION_MAP/ LOAD_OCCUPANCY_GRID = 2

Loading both would be possible as:
LOAD_OCCUPANCY_GRID | LOAD_LOCALIZATION_MAP

tango_ros_common/tango_ros_messages/srv/SaveMap.srv
```diff
@@ -1,6 +1,11 @@
...
+int8 ONLY_NAV_MAP=2
+int8 request
 string map_name
 ---
 string message
-string map_name
-string map_uuid
+string localization_map_name
+string localization_map_uuid
+string navigation_map_name
```
mcopejans: Will the occupancy grid map always be used for navigation or could it also be used to e.g. represent the walls in a game environment or to measure the surface of the walls in a home? In this case, I would rename all occurences to occupancy_grid.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
PerrineAguiar: The function is already called in `onTangoStatus`. But it is not sufficient, the buttons also need to be updated in `onConfigurationChanged`.

PerrineAguiar:
@mcopejans Thanks for the review, I think I addressed all your comments.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -364,6 +387,40 @@ public void onClick(View view) {
         }
     }
 
+    public void onClickOkLoadOccupancyGridDialog(final String occupancyGridName) {
+        if (occupancyGridName == null || occupancyGridName.isEmpty()) {
+            Log.e(TAG, "Name is null or empty, unable to load the occupancy grid");
+            displayToastMessage(R.string.name_error);
```
mcopejans: Instead of showing a toast message which can easily be missed, I would display the dialog again. Possibly show extra text in red that previous map could not be loaded. Cancel button is still there if a user would like to cancel the operation.

Alternative: just check fail here or raise an exception because it should never happen.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -364,6 +387,40 @@ public void onClick(View view) {
...
+            return;
+        }
+        new AsyncTask<Void, Void, Void>() {
+            @Override
+            protected Void doInBackground(Void... params) {
+                mTangoServiceClientNode.callLoadOccupancyGridService(occupancyGridName);
+                return null;
+            }
+        }.execute();
+        mLoadOccupancyGridButton.setEnabled(false);
```
mcopejans: Nit: first disable the button, then call the service. If something goes wrong while calling the service, the button should already be disabled.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
mcopejans: This function is called: `restartTango`, so in this case `onTangoStatus` will be called.
Could you move the function call `updateLoadAndSaveMapButtons` to `onConfigurationChanged` instead?

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/LoadOccupancyGridDialog.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public View onCreateView(LayoutInflater inflator, ViewGroup container,
+                             Bundle savedInstanceState) {
+        View dialogView = inflator.inflate(R.layout.dialog_load_occupancy_grid, null);
+        getDialog().setTitle(R.string.load_occupancy_grid_dialog_title);
+        mOkButton = (Button) dialogView.findViewById(R.id.load_occupancy_grid_ok);
+        mOkButton.setOnClickListener(this);
+        dialogView.findViewById(R.id.load_occupancy_grid_cancel).setOnClickListener(this);
+        mNameEditText = (EditText) dialogView.findViewById(R.id.load_occupancy_grid_name);
+        mNameEditText.addTextChangedListener(new TextWatcher() {
```
mcopejans: Can you replace the name edit text field by a list with occupancy grids that are available to load?

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/android/SaveMapDialog.java
```diff
@@ -42,34 +47,54 @@
...
-        getDialog().setTitle("Enter the name of the map");
-        mNameEditText = (EditText) dialogView.findViewById(R.id.map_name);
-        dialogView.findViewById(R.id.save_map_ok).setOnClickListener(this);
+        getDialog().setTitle(R.string.save_map_dialog_title);
+        mOkButton = (Button) dialogView.findViewById(R.id.save_map_ok);
+        mOkButton.setOnClickListener(this);
         dialogView.findViewById(R.id.save_map_cancel).setOnClickListener(this);
-        setCancelable(false);
+        mNameEditText = (EditText) dialogView.findViewById(R.id.map_name);
+        mNameEditText.addTextChangedListener(new TextWatcher() {
```
mcopejans: I think it would be helpful for a user to generate a default map name based on e.g. the current date and time: `map_20170704_142400`. Only if this default name is intentionally erased, the Ok button is disabled.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
PerrineAguiar: The functionality to display the dialog again was added in a previous commit. 
Also, the OK button is now disabled if the text field is empty, so you are right it should never happen. I will add an assert here.

PerrineAguiar:
@mcopejans Thanks again, I addressed your comments.

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -288,7 +296,19 @@ private void showSaveMapDialog() {
         FragmentManager manager = getFragmentManager();
         SaveMapDialog saveMapDialog = new SaveMapDialog();
         saveMapDialog.setStyle(DialogFragment.STYLE_NORMAL, R.style.CustomDialog);
-        saveMapDialog.show(manager, "MapNameDialog");
+        saveMapDialog.show(manager, "SaveMapDialog");
+    }
+
+    private void showLoadOccupancyGridDialog(boolean firstTry, java.util.ArrayList<java.lang.String> nameList) {
```
mcopejans: Why is it necessary to make a distinction between the first try and other, later tries?

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -316,15 +336,30 @@ public void onClick(View view) {
...
+                File occupancyGridDirectory = new File(directory);
+                if (occupancyGridDirectory != null && occupancyGridDirectory.isDirectory()) {
+                    File[] files = occupancyGridDirectory.listFiles();
+                    for (File file : files) {
+                        if (FilenameUtils.getExtension(file.getName()).equals("yaml")) {
+                            mOccupancyGridNameList.add(FilenameUtils.removeExtension(file.getName()));
+                        }
+                    }
+                }
+                showLoadOccupancyGridDialog(true, mOccupancyGridNameList);
```
mcopejans: showDialog(/* first_try */ true, ...)

TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
PerrineAguiar: If it failed to load the previous map, we want  to show an extra message in the dialog saying "Unable to load the occupancy grid, please try again with a different one.".

PerrineAguiar:
@mcopejans PR merged! It was a big one, thanks again for your reviews!

PR #335, opened 2017-06-29
Feature/use space layer with threshold for occupancy grid

PerrineAguiar:
This PR changes the code to make the occupancy grid from the space layer instead of the obstacle layer. 
This produces navigation map with less "fake points". 
Moreover, since the space layer is not binarized, it adds a threshold parameter to decide if a pixel value corresponds to a free or occupied cell.

Before this PR:
image

After this PR (threshold = 128):
image

After this PR (threshold = 110):
image

After this PR (threshold = 70):
image

### DIFF
--- a/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
+++ b/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
@@ -28,9 +28,9 @@
 
 namespace tango_ros_conversions_helper {
 const int NUMBER_OF_FIELDS_IN_POINT_CLOUD = 4;
-constexpr int FREE_CELL = 0;
-constexpr int OCCUPIED_CELL = 100;
-constexpr int UNKNOWN_CELL = -1;
+constexpr int8_t FREE_CELL = 0;
+constexpr int8_t OCCUPIED_CELL = 100;
+constexpr int8_t UNKNOWN_CELL = -1;
 
 // Converts a TangoPoseData to a geometry_msgs::TransformStamped.
 // @param pose, TangoPoseData to convert.
--- a/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
+++ b/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
@@ -134,7 +134,7 @@ void toMeshMarker(const Tango3DR_GridIndex& grid_index,
 // @param occupancy_grid, the output nav_msgs::OccupancyGrid.
 void toOccupancyGrid(const Tango3DR_ImageBuffer& image_grid,
                      const Tango3DR_Vector2& origin,
-                     double time_offset, double resolution, int threshold,
+                     double time_offset, double resolution, uint8_t threshold,
                      nav_msgs::OccupancyGrid* occupancy_grid);
 
 // Converts Tango3DR_Vector3 to geometry_msgs::Point.
--- a/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
+++ b/tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
@@ -127,10 +127,14 @@ void toMeshMarker(const Tango3DR_GridIndex& grid_index,
 // @param origin, position of the top left pixel in
 /// world coordinates (x: right, y: up).
 // @param resolution the grid resolution (m/cell).
+// @param threshold Threshold to decide if a pixel value corresponds to a free
+// or occupied cell. Should be between 0 and 255.
+// Pixel value <= threshold --> cell is free.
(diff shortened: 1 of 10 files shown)


tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_node.h
```diff
@@ -275,6 +277,7 @@ class TangoRosNode : public ::nodelet::Nodelet {
   Tango3DR_Pose last_camera_color_pose_;
   Tango3DR_CameraCalibration t3dr_color_camera_intrinsics_;
   double t3dr_resolution_;
+  int t3dr_occupancy_grid_threshold_ = TANGO_3DR_OCCUPANCY_GRID_DEFAULT_THRESHOLD;
```
smits: For consistency reasons I would not assign the default value here just yet.

tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
```diff
@@ -127,10 +127,14 @@ void toMeshMarker(const Tango3DR_GridIndex& grid_index,
...
 // @param resolution the grid resolution (m/cell).
+// @param threshold Threshold to decide if a pixel value corresponds to a free
+// or occupied cell. Should be between 0 and 255.
+// Pixel value <= threshold --> cell is free.
+// Pixel value > threshold --> cell is occupied.
 // @param occupancy_grid, the output nav_msgs::OccupancyGrid.
 void toOccupancyGrid(const Tango3DR_ImageBuffer& image_grid,
                      const Tango3DR_Vector2& origin,
-                     double time_offset, double resolution,
+                     double time_offset, double resolution, int threshold,
```
smits: Since the values of the layer are uint8, it makes sense to make the threshold parameter a uint8_t too?

PerrineAguiar:
@smits Thanks for your review, I addressed your comments.

tango_ros_common/tango_ros_native/src/tango_ros_conversions_helper.cpp
```diff
@@ -277,13 +277,14 @@ void toOccupancyGrid(const Tango3DR_ImageBuffer& image_grid,
...
       // flipped around the x axis.
-      int value = static_cast<int>(image_grid.data[j + (image_grid.height - i - 1) * image_grid.width]);
-      if (value == 1) {
-        occupancy_grid->data.push_back(OCCUPIED_CELL);
-      } else if (value == 255) {
-        occupancy_grid->data.push_back(FREE_CELL);
-      } else if (value == 128) {
+      int value = static_cast<int>(
+          image_grid.data[j + (image_grid.height - i - 1) * image_grid.width]);
+      if (value == 128) {
```
smits: Is there a reason why we cast this to an int instead of a uint8_t?

PerrineAguiar:
@smits Thanks, I addressed your comments.

tango_ros_common/tango_ros_native/src/tango_ros_conversions_helper.cpp
```diff
@@ -277,13 +277,13 @@ void toOccupancyGrid(const Tango3DR_ImageBuffer& image_grid,
...
-      } else if (value == 255) {
-        occupancy_grid->data.push_back(FREE_CELL);
-      } else if (value == 128) {
+      uint8_t value = image_grid.data[j + (image_grid.height - i - 1) * image_grid.width];
+      if (value == 128) {
         occupancy_grid->data.push_back(UNKNOWN_CELL);
+      } else if (value <= threshold) {
+        occupancy_grid->data.push_back(FREE_CELL);
+      } else if (value > threshold) {
+        occupancy_grid->data.push_back(OCCUPIED_CELL);
```
smits: I guess we better make UNKOWN, OCCUPIED and FREE cell uint8_t's as well

PerrineAguiar: Done. I changed their type to in8_t, since it can be negative.

PerrineAguiar:
@smits I addressed your last comment.

tango_ros_common/tango_ros_native/include/tango_ros_native/tango_ros_conversions_helper.h
```diff
@@ -28,9 +28,9 @@
 
 namespace tango_ros_conversions_helper {
 const int NUMBER_OF_FIELDS_IN_POINT_CLOUD = 4;
-constexpr int FREE_CELL = 0;
-constexpr int OCCUPIED_CELL = 100;
-constexpr int UNKNOWN_CELL = -1;
+constexpr int8_t FREE_CELL = 0;
+constexpr int8_t OCCUPIED_CELL = 100;
+constexpr int8_t UNKNOWN_CELL = -1;
```
smits: We could still make this a uint8 and use 255 instead?

PR #338, opened 2017-07-05
Fix/bug adf not saved

### LINKED ISSUE #334: ADF saving does not always get initiated after clicking on the save button
No trace on the Tango core of a saving request.

PerrineAguiar:
This PR fixes #334.
When calling `TangoService_saveAreaDescription` the sensors are stopped before actually saving the ADF. Therefore the call to `TangoService_getPoseAtTime` was not returning which was causing calls to the `SaveMap` service to hang.

### DIFF
--- a/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
+++ b/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
@@ -661,7 +661,6 @@ void TangoRosNode::PublishStaticTransforms() {
 }
 
 void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
-  //LOG(INFO) << "OnPoseAvailable";
   if (publish_pose_on_tf_ ||
       start_of_service_T_device_publisher_.getNumSubscribers() > 0 ||
       area_description_T_start_of_service_publisher_.getNumSubscribers() > 0) {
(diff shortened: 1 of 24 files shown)


tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -668,16 +668,18 @@ void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
...
-        pair.base = TANGO_COORDINATE_FRAME_AREA_DESCRIPTION;
-        pair.target = TANGO_COORDINATE_FRAME_START_OF_SERVICE;
-        TangoPoseData area_description_T_start_of_service;
-        TangoService_getPoseAtTime(0.0, pair, &area_description_T_start_of_service);
-        if (area_description_T_start_of_service.status_code == TANGO_POSE_VALID) {
-          tango_ros_conversions_helper::toTransformStamped(area_description_T_start_of_service,
-                             time_offset_, &area_description_T_start_of_service_);
+        tango_ros_conversions_helper::toTransformStamped(
+            *pose, time_offset_, &start_of_service_T_device_);
+        if (tango_data_available_) {
```
mcopejans: This solution couples the save map function with onPoseAvailable, which is quite unexpected. Either add a big comment why this is necessary or consider the following alternative solution which avoids a manual call to getPoseAtTime. Register multiple frame pairs so getPoseAtTime is called for sos_T_adf and sos_T_device.
```
void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
  if (base == area description && target == device) {
      // Save start_of_service_T_device
  } else if (base == area description && target == start of service
     // Save area_description_T_start_of_service_
  }
)
```

PerrineAguiar:
@mcopejans Thanks for your review. I went for registering adf_sos as well instead of using `TangoService_getPoseAtTime`.

PR #347, opened 2017-07-11
Feature/add uuid in load service response

### LINKED ISSUE #344: return uuid in occupancy grid loading service

PerrineAguiar:
This PR adds the ADF uuid as part of the response from the LoadOccupancyGrid service.

Fixes #344.


PerrineAguiar:
It can be useful for a client to know the corresponding map uuid, e.g. to load the corresponding ADF afterwards. Otherwise the client would need to get the uuid directly from the yaml file, which is not straightforward since the file is located on the device.

PerrineAguiar:
@smits ptal

PerrineAguiar:
@smits Thanks for your review.

PR #349, opened 2017-07-12
Save occupancy grid wrt area description frame

### LINKED ISSUE #346: Save occupancy map wrt adf

PerrineAguiar:
With this PR the occupancy grid is saved wrt the area description frame.

Fixes #346.


tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -745,30 +760,26 @@ void TangoRosNode::PublishStaticTransforms() {
 }
 
 void TangoRosNode::OnPoseAvailable(const TangoPoseData* pose) {
-  if (publish_pose_on_tf_ ||
-      start_of_service_T_device_publisher_.getNumSubscribers() > 0 ||
-      area_description_T_start_of_service_publisher_.getNumSubscribers() > 0) {
```
smits: Can you explain me why these checks are not required anymore.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -905,9 +916,7 @@ void TangoRosNode::StopPublishing() {
   if (run_threads_) {
     run_threads_ = false;
     if (device_pose_thread_.publish_thread.joinable()) {
-      if (!tango_data_available_ || (!publish_pose_on_tf_ &&
-          start_of_service_T_device_publisher_.getNumSubscribers() <= 0 &&
-          area_description_T_start_of_service_publisher_.getNumSubscribers() <= 0)) {
```
smits: Same here

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: To be able to express the origin of the map in the area description frame we need to know the transform `area_description_T_start_of_service`, which is filled in only after these checks. I removed them so that we are sure that `area_description_T_start_of_service` is always known.

These checks were here to avoid slowing down the callback if no one is using the pose data. But I don't think it is really necessary for the pose callback.
Consequently I also removed the corresponding checks in the `StopPublishing` method, which were there to unlock the publishing thread in case we can not reach the `notify_all()` call inside the Tango callbacks.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: See comment above.

PerrineAguiar:
It is possible to build the occupancy map directly wrt the adf frame, but this would require to transform the origin coordinates into the adf frame for each new image.

PerrineAguiar:
@smits ptal

PerrineAguiar:
@smits Thanks for reviewing.

PR #355, opened 2017-07-13
Fix/avoid empty occupancy grid when no subscriber

### LINKED ISSUE #345: avoid empty occupancy map if no subscriber

PerrineAguiar:
This PR adds two ROS parameters to enable/disable mesh and occupancy grid. That way we avoid relying on the number of subscribers to extract and publish them.

Fixes #345.

### DIFF
--- a/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
+++ b/tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
@@ -1110,14 +1110,12 @@ void TangoRosNode::PublishMesh() {
             &t3dr_updated_indices);
       }
       // Publish Tango mesh as visualization marker.
-      if (enable_3dr_mesh_) {
+      if (enable_3dr_mesh_ && mesh_marker_publisher_.getNumSubscribers() > 0) {
         visualization_msgs::MarkerArray mesh_marker_array;
         tango_3d_reconstruction_helper::ExtractMeshAndConvertToMarkerArray(
             t3dr_context_, t3dr_updated_indices, time_offset_,
             start_of_service_frame_id_, &mesh_marker_array);
-        if (mesh_marker_publisher_.getNumSubscribers() > 0) {
-          mesh_marker_publisher_.publish(mesh_marker_array);
-        }
+        mesh_marker_publisher_.publish(mesh_marker_array);
         Tango3DR_Status result = Tango3DR_GridIndexArray_destroy(
             &t3dr_updated_indices);
         if (result != TANGO_3DR_SUCCESS) {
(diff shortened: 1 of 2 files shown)


PerrineAguiar:
@smits I added the check around the publish call.

PerrineAguiar:
@smits ptal

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -1120,11 +1129,12 @@ void TangoRosNode::PublishMesh() {
...
       }
       // Publish Tango mesh as occupancy grid.
-      if (occupancy_grid_publisher_.getNumSubscribers() > 0) {
+      if (enable_3dr_occupancy_grid_) {
         occupancy_grid_.data.clear();
         if (tango_3d_reconstruction_helper::ExtractFloorPlanImageAndConvertToOccupancyGrid(
             t3dr_context_, time_offset_, start_of_service_frame_id_,
-            t3dr_resolution_, t3dr_occupancy_grid_threshold_, &occupancy_grid_))
+            t3dr_resolution_, t3dr_occupancy_grid_threshold_, &occupancy_grid_) &&
+            occupancy_grid_publisher_.getNumSubscribers() > 0)
```
smits: Can we move the getNumSubscribers check around the Extract call, I guess if we are not going to publish it we don't need to extract the floorplan image?

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -1103,12 +1110,14 @@ void TangoRosNode::PublishMesh() {
...
       }
       // Publish Tango mesh as visualization marker.
-      if (mesh_marker_publisher_.getNumSubscribers() > 0) {
+      if (enable_3dr_mesh_) {
         visualization_msgs::MarkerArray mesh_marker_array;
         tango_3d_reconstruction_helper::ExtractMeshAndConvertToMarkerArray(
             t3dr_context_, t3dr_updated_indices, time_offset_,
             start_of_service_frame_id_, &mesh_marker_array);
-        mesh_marker_publisher_.publish(mesh_marker_array);
+        if (mesh_marker_publisher_.getNumSubscribers() > 0) {
```
smits: Can we move this check around the Extract function? No need to extract anything as long as we are not publishing?

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: Yes we do. It can happen that there is no subscriber for the occupancy grid but we still want to extract it to be able to save it.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
PerrineAguiar: I agree that the case of the mesh is different since we don't need the data somewhere else. I will make the change.

PR #356, opened 2017-07-20
New release

PerrineAguiar:
This PR increases version numbers for the new release.


PerrineAguiar:
@smits ptal

PerrineAguiar:
Because we changed the SaveMap service request/response, which is not backward compatible.

PR #358, opened 2017-07-28
Publish area_dexcription to sos tf as static

PerrineAguiar:
Tango updates the transform between area_description and start_of_service only every 0.1 s. This can cause some applications to work slower than expected.

For example, in the case of the ROS navigation stack, the control loop of the [move_base](http://wiki.ros.org/move_base) package takes ~0.5 s. The reason is that it uses the [base_local_planner](http://wiki.ros.org/base_local_planner?distro=lunar) which makes several calls to [waitForTransform](http://docs.ros.org/jade/api/tf/html/c++/classtf_1_1Transformer.html#a72dc26fe7bfcb9585123309e092e5c83). Each call is blocking for between 0.1 to 0.2 s when waiting for the tf between area_description and start_of_service.

To avoid this, this PR replaces the use of tf by tf_static to publish the transform between area_description and start_of_service.


PerrineAguiar:
@dvanthienen @mcopejans Thanks

PR #366, opened 2017-08-01
Create occupancy grid directory if it does not exist

### LINKED ISSUE #364: Create occ map directory automatically if does not exist

PerrineAguiar:
Fixes #364.

### DIFF
--- a/tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
+++ b/tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
@@ -27,20 +27,16 @@
 #include <yaml-cpp/yaml.h>
 
 namespace {
-void AddTrailingSlashToDirectoryPathIfNeeded(std::string& directory_path) {
-  if (!directory_path.empty() && directory_path.back() != '/') {
-    directory_path += '/';
+void AddTrailingSlashToDirectoryPathIfNeeded(std::string* directory_path) {
+  if (!directory_path->empty() && directory_path->back() != '/') {
+    (*directory_path) += '/';
   }
 }
 bool DirectoryPathExists(const std::string& path) {
   struct stat st;
-  if (stat(path.c_str(), &st) == 0 && (st.st_mode & S_IFDIR)) {
-    return true;
-  } else {
-    return false;
-  }
+  return (stat(path.c_str(), &st) == 0 && (st.st_mode & S_IFDIR));
 }
-int MakeDirectoryPath(const std::string& input_path, mode_t mode) {
+bool MakeDirectoryPath(const std::string& input_path, mode_t mode) {
   if (input_path.empty()) {
     // Path is empty, there is nothing to create so we return success.
     return 0;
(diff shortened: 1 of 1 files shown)


tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
```diff
@@ -27,9 +29,47 @@
 namespace {
 void AddTrailingSlashToDirectoryPathIfNeeded(std::string& directory_path) {
   if (!directory_path.empty() && directory_path.back() != '/') {
-    directory_path += "/";
+    directory_path += '/';
   }
 }
+bool DirectoryPathExists(const std::string& path) {
+  struct stat st;
+  if (stat(path.c_str(), &st) == 0 && (st.st_mode & S_IFDIR)) {
```
mcopejans: Directly return content of if statement instead if true false

tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
```diff
@@ -27,9 +29,47 @@
...
+    return 0;
+  }
+
+  size_t previous_slash_pos = 0;
+  size_t current_slash_pos;
+  std::string dir;
+  int make_dir_status = 0;
+
+  std::string path = input_path;
+  CHECK(!path.empty());
```
mcopejans: I think it is ok to return 0 for success in this case instead of check failing. Empty path = no directory = directory already exists

tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
```diff
@@ -27,9 +29,47 @@
...
+  AddTrailingSlashToDirectoryPathIfNeeded(path);
+
+  while ((current_slash_pos = path.find_first_of('/', previous_slash_pos)) !=
+         std::string::npos) {
+    dir = path.substr(0, current_slash_pos++);
+    previous_slash_pos = current_slash_pos;
+    if (dir.empty())
```
mcopejans: Add curly brackets for multiline expression. Also combine if statement here and below with OR.

tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
```diff
@@ -45,6 +85,13 @@ bool SaveOccupancyGridDataToPgmFile(
     const nav_msgs::OccupancyGrid& occupancy_grid) {
   std::string map_directory_with_trailing_slash = map_directory;
```
mcopejans: This can be const ref.

tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
```diff
@@ -45,6 +85,13 @@ bool SaveOccupancyGridDataToPgmFile(
...
   AddTrailingSlashToDirectoryPathIfNeeded(map_directory_with_trailing_slash);
+  if (!DirectoryPathExists(map_directory_with_trailing_slash)) {
+    LOG(INFO) << "Directory " << map_directory_with_trailing_slash << " does not exist, creating it now.";
+    if (MakeDirectoryPath(map_directory_with_trailing_slash, S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH) != 0) {
```
mcopejans: Easier if MakeDirectoryPath returns a Boolean. It is not directly clear from outside the function what the integers mean in the specific implementation if this function.

tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
```diff
@@ -27,9 +29,47 @@
 namespace {
 void AddTrailingSlashToDirectoryPathIfNeeded(std::string& directory_path) {
   if (!directory_path.empty() && directory_path.back() != '/') {
-    directory_path += "/";
+    directory_path += '/';
   }
 }
+bool DirectoryPathExists(const std::string& path) {
```
mcopejans: These functions are very generic so deserve their own file: file_io.cpp. This is only an opinion so ok to move them later, if they are reused elsewhere in the code.

tango_ros_common/tango_ros_native/src/occupancy_grid_file_io.cpp
PerrineAguiar: I could not use const ref here because `AddTrailingSlashToDirectoryPathIfNeeded` had a non const reference argument. I changed it to use a pointer instead to avoid confusion.

PerrineAguiar:
I will merge this as soon as the jenkins job is finished.

PR #367, opened 2017-08-01
Fix/better default for occupancy grid threshold

### LINKED ISSUE #365: Choose better default value for occ grid threshold

PerrineAguiar:
This PR changes the default value of the occupancy grid threshold to 180.
It was set to 128, which does not give good result.

Fixes #365.


mcopejans:
LGTM
Is a 'good' value for this threshold device dependent? Is this a good value for the tablet, Lenovo and Asus devices?

PerrineAguiar:
This was tested only on Lenovo, so I am not sure if this is device dependent.
I am merging this PR for now since Lenovo is the most used currently.

PR #371, opened 2017-08-10
Add api level parameter to avoid connecting to fisheye on Asus

### LINKED ISSUE #314: Failed to connect to Tango on Asus ZenAR
I'm getting the following error:
```
06-14 10:44:02.310 15658 15788 I tango_client_api: TangoErrorType TangoService_connectOnFrameAvailable(TangoCameraId, void*, void (*)(void*, TangoCameraId, const TangoImageBuffer*)): Connecting callbacks...
06-14 10:44:02.310 15658 15788 E tango_client_api: Cannot register onFrameAvailable callback when onImageAvailable already registered!
06-14 10:44:02.310 15658 15788 E native  : tango_ros_node.cpp:530 TangoRosNode::TangoConnect(), TangoService_connectOnFrameAvailable TANGO_CAMERA_FISHEYE error: -2
```
I don't see us registering a callback onImageAvailable though.  This happens with both the Tango G and I release and with both our master and our latest version on the PlayStore.

PerrineAguiar:
`TangoService_connectOnFrameAvailable` is not supported on Asus phones for the fisheye camera. This was causing the app to fail connecting to Tango on Asus phones.

To fix this, this PR exposes the API level of the device on a ROS parameter and decide to connect to 'onFrameAvailable' or not  (Asus phone has API level 24 while Lenovo has 23).

Fixes #314.

### DIFF
--- a/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
+++ b/TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
@@ -27,6 +27,7 @@ import android.net.Uri;
 import android.content.res.Configuration;
 import android.net.wifi.WifiManager;
 import android.os.AsyncTask;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.StrictMode;
 import android.preference.PreferenceManager;
@@ -60,9 +61,14 @@ import org.ros.node.NodeMainExecutor;
 import java.io.File;
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import eu.intermodalics.nodelet_manager.TangoNodeletManager;
 import eu.intermodalics.nodelet_manager.TangoInitializationHelper;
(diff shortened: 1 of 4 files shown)


TangoRosStreamer/app/src/main/java/eu/intermodalics/tango_ros_streamer/activities/RunningActivity.java
```diff
@@ -747,7 +753,35 @@ protected void init(NodeMainExecutor nodeMainExecutor) {
...
+                        public Boolean call() throws Exception{
+                            while (connectedNode.getParameterTree() == null) {
+                                try {
+                                    Thread.sleep(200);
+                                } catch (InterruptedException e) {
+                                    e.printStackTrace();
+                                    return false;
+                                }
+                            }
+                            mParameterNode.setIntParam("android_api_level", Build.VERSION.SDK_INT);
```
mcopejans: Can you move the code to set a parameter to the `onStart` method of mParameterNode? This seems a more simple solution, but maybe I am missing something.

PerrineAguiar:
@mcopejans Thanks, I addressed your comment.

tango_ros_common/tango_ros_native/src/tango_ros_node.cpp
```diff
@@ -419,6 +410,21 @@ TangoErrorType TangoRosNode::OnTangoServiceConnected() {
...
+  if (fisheye_camera_available_) {
+    try {
+        fisheye_camera_publisher_ =
+            image_transport_->advertiseCamera(FISHEYE_IMAGE_TOPIC_NAME,
+                                              queue_size, latching);
+        fisheye_rectified_image_publisher_ =
+            image_transport_->advertise(FISHEYE_RECTIFIED_IMAGE_TOPIC_NAME,
+                                       queue_size, latching);
+      } catch (const image_transport::Exception& e) {
+        LOG(ERROR) << "Error while creating fisheye image transport publishers" << e.what();
```
mcopejans: Should this also set fisheye_camera_available_ to false?

PerrineAguiar: I don't think so, this boolean indicates if the camera is available on the Tango side.

mcopejans: I understand, however the point is that no fisheye camera images should published in the end so this is why I suggested this. As long as the code doesn't segfault when this error occurs, it's fine for me.

PerrineAguiar: I see, in this case we could return an error if this exception occurs. What do you think?

PR #381, opened 2017-11-14
Release tango_ros_messages as binary packages into ROS indigo and kinetic

meyerj:
This is required to release `tango_ros_messages` version 1.3.1 using bloom.

All other packages cannot easily be submitted for binary releases to the ROS build farm because the third-party dependencies, including the Tango Client API, cannot be released without permission and are not available for all ROS target platforms. But the message package is already useful to communicate with the TangoRosStreamer app running on an Android device.

### DIFF
--- a/tango_ros_common/tango_ros_messages/package.xml
+++ b/tango_ros_common/tango_ros_messages/package.xml
@@ -12,7 +12,6 @@
   <build_depend>message_generation</build_depend>
   <build_depend>std_msgs</build_depend>
 
-  <run_depend>message_generation</run_depend>
   <run_depend>message_runtime</run_depend>
   <run_depend>std_msgs</run_depend>
 


meyerj:
I created a [bloom](http://wiki.ros.org/bloom) release repository at https://github.com/Intermodalics/tango_ros-release with a copy of version 1.3.1 from this branch and initial releases into ROS indigo and kinetic. All went well and we could submit a PR to https://github.com/ros/rosdistro to trigger the build farm?

PerrineAguiar:
@meyerj Looks good but won't this conflict with the rosjava version of the tango_ros_messages package (see [here](https://github.com/rosjava/rosjava_mvn_repo/tree/master/org/ros/rosjava_messages/tango_ros_messages)), which is at 2.0.0?

meyerj:
Probably yes. In this case we have to keep the version number `2.0.0`, which is not a problem as long as all other packages are ignored. But normally bloom expects that all packages within the same repository have the same version, and that version should correspond to the git tag name somehow.

I will revert the first commit and only keep the second and do another release for version 2.0.0.

meyerj:
Done:
- https://github.com/ros/rosdistro/pull/16550
- https://github.com/ros/rosdistro/pull/16551

